PR #411, opened 2014-04-01
E1406 & E1414. Testing classes related to teams

purrrfecto:
I have heard that our first attempt of pull request has some problem, so I try to commit the file from my version.


yangsong8:
They use the same pull request for both oss and final project.

PR #444, opened 2014-10-28
E1467: Refactoring Leaderboard model

ravishroshan:
The pull request is for a refactored implementation of Leaderboard implementation. Our team has reduced the the database calls, loops and redundant storage and computation. We have refactored many files related to Leaderboard implementation leading to enhancement of overall complexity of the feature.
Following is a snapshot of complexity improvement as measured by code climate.

comparing rails4 expertiza code climate


Winbobob:
See final project same semester by same team #469

PR #469, opened 2014-12-05
E1472: Connect changes to score model with changes to score views

maxlpy:
What we need to do:
Remove the get_ and set_ accessor names; just use Ruby accessors.
Change camel-case to underscores to match the Ruby convention.
In grades/_participant.html.erb, don’t create local variables in the view; rather, create an object that has max, min, and avg fields for each kind of reviews. Try to avoid the need to check review type; instead, just pass the kind of review when referring to a max, min, or avg field and use polymorphism.
Also, you shouldn't have to check for the existence topic in the view; if it is null, a null string should just be returned.
There are lots of calculations in the views, which should be moved to the model.

What we have done:
We modified the camel-case variables in views/grade to Ruby convention by replacing them with under_score format while keeping the camel-case in JavaScript, as in JavaScript the camel-case should be its convention.
We set up new data structure Rscore to encapsulate the score and reduce the local variables.
We analysed the system and found out that polymorphism for Rscore is not necessary, encapsulation will handle all the usages.
We moved the calculations in the views into model file,assignment_participants.rb, which follows the single responsibility principle and DRY principle.
After confirmation with Professor and TAs, we no longer need to check the existence of the topic.
We modified the response_map to expedite the rendering speed of showing scores.


app/models/assignment_participant.rb
```diff
@@ -149,52 +149,51 @@ def get_reviewers
 
     reviewers
   end
-  alias_method :reviewers, :get_reviewers
```
akofink: Maybe we should keep these aliases just in case other code uses the get_x methods.

Issue #504, opened 2015-03-01
If first time new-user creation occur some unsatisfaction, cannot create new users any more


Winbobob:
![expertiza_bug](https://cloud.githubusercontent.com/assets/7702035/6432340/6c5a4ec4-c01b-11e4-9a16-b041413bb076.png)

Winbobob:
The only way to create new users again is clear your browser's cache.

Winbobob:
In Rails 4.1. these are no longer deprecation warnings if ```with_scope```, but errors.

Winbobob:
This bug has been fixed by @yatish27. And the way to fix this bug is to update the ```Gemfile``` to newest version.

yatish27:
Not just updating the gem. the authlogic expertiza needed some changes. I fixed that in the expertiza/authlogic repo.

PR #505, opened 2015-03-16
E1506: Refactoring, testing and new features related to “users”

Winbobob:
<b>What needed to be done:</b>
Classes involved: users_controller.rb
-   Find the un-called methods if any and delete them. <b>[done]</b>
-  Change the Rails 2 syntax to Rails 4 style.
-  Refactor users_controller.rb
  -  Change the white space for the second harf of this file, starts at “def edit”. <b>[done]</b>
  -  Separate the “paginate_list” method into two methods. The search method should be in model and the paginating method should be in the controller.  <b>[done]</b>
-   New feature: delete users
  - A user can be deleted if (s)he has not participated in an assignment. <b>[done]</b>
  - If the user is participating in an assignment, the system will ask, “User is participating in k assignments.  Delete as a participant in these assignment(s)?”  <b>[done]</b>
 - If the user has submitted or reviewed in any of these assignments, the system will say the user cannot be deleted, but offer to rename the user account to <current_account_name>_hidden.
   -  rename (javascript calling update method in users_controller.rb) <b>[done]</b>
    - different users have different delete methods.  <b>[done]</b>
 - If the person trying to delete does not want to rename the account, the system will just say that the user can’t be deleted. <b>[done]</b>
 -  Write tests (with Rspec) for this feature.<b>[Done (users without relationships)]</b>
-  Testing feature: search for users
 -  In rails 4 branch, admins can search for the users with 1) users’ login names 2) users’ last or first names and 3) users’ emails. Please write tests (with Rspec) for this feature <b>[done]</b>


spec/features/users_spec.rb
```diff
@@ -1,21 +1,59 @@
 require 'rails_helper'
 include LogInHelper
 
-feature 'Users' do
-  before(:each) do
+feature 'Method 1: Instructor search a user' do
+  before(:all) do
```
Winbobob: If #504 issue can be fixed, we can use ```before(:each)``` instead of ```before(:all)```. Now we cannot run ```student.save``` twice, because it will generate new users twice and occur an error describe in #504.

app/views/users/show.html.erb
```diff
@@ -20,8 +20,13 @@
 <% unless current_user_role? == "Student" %>
   <hr/>
   <p><%= link_to 'Edit', :action => 'edit', :id => @user %> |
-    <%= link_to 'Delete', {:action => 'destroy', :id => @user},
-    :confirm => 'Delete this user?', :method => :post %>
+<!--This user cannot be deleted directly-->
+    <% if (@assignment_participant_num != 0)%>
```
Winbobob: Still need modify this if condition. We want to use the result of ```Response_mapping``` database, but we find it a little werid: we create a new user and do nothing, we can find corresponding recode in ```Response_mapping``` database as reviewer.

yangsong8:
Hi Zhewei, we are going to merge this pull request. Before we merge let me just give you a checklist on the issues we expect you to fix before merging.

- [ ] Merge conflicts in the repo. (some undeleted "<<<HEAD")
- [ ]Bug: we tried to create a user without setting password. Then we impersonate the new user and set password, however, we got a bug when we try to save the user's profile.
We are not sure if the test for reviewing is implemented correctly.  The user has reviewed iff (s)he has created any Response objects.  
- [x]On the users list page, there should be a checkbox that says, Show hidden users? (default unchecked).  Also, before a user is deleted, the code needs to check to be sure it is not deleting the super_admin!  Or actually, not just the super_admin, but the user must not be allowed to delete any user not created by him/her or one of his subordinate users.

Winbobob:
Hi Yang, we have already finished ```1``` and ```2```. About ```4```, I recommend waiting Ravi to fix "impersonate" functionality. Because I think "delete user" and "impersonate"  should have same limits of authority.

Winbobob:
And about ```2```, we will fix it as soon as possible.

yangsong8:
You can check those boxes after you finish them. Thanks.

Winbobob:
Hi Yang, I have already done manual test. And I find same error as #504, but when I go back to previous page and click ```save``` again. It works well.

Winbobob:
I think updating user profile has the same error of creating new user twice without clean the cache. I am so sorry that I am not able to fix this #504 bug right now.

Winbobob:
However, I think merging our OSS project is fine, because it has no confilct with ```rails4``` version.

PR #510, opened 2015-03-17
E1501 Refactoring assignment.rb

yatish27:


yangsong8:
According to our record your team did not set a test server for the code you have changed. It would be great if you can demo your change on the coming Expertiza meeting.

yatish27:
We have setup the test server. We sent mail to the TAs regarding it @nikhil was looking into it.

PR #531, opened 2015-04-02
E1509 Refactoring SignUpController and SignUpSheetController

jli53:
Here are our modifications for OSS project topic E1509


yangsong8:
Hi, did you make any change on the DB? Thanks.

jli53:
Hi Yang,

We did not change anything on DB. Is there anything wrong about it?

Thanks,
Jiaming

yangsong8:
It is about the changes on db schema. If you did not change the db, we can
ignore the changed on that file when we merge your project.

jli53:
I checked the schematic.rb again. We did not mean to change anything on
this file. But it seems that when we delete somethhing from a file in
"view" folder, this chema file also get changed.

yangsong8:
I mean the file: db/schema.rb. Your team committed this file and changed several hundreds on lines, which made this pull request contains more changed lines than we expected. This is the reason I asked if you guys made any change on the DB. If your answer is "no", this means the db/schema.rb is changed by Rails. So we will not worry about this file when we merge your changes.

By the way, in your final project, you do not need to commit db/schema.rb unless you really change the DB.

Thank you for your time.

jli53:
Our answer is "no". We didn't change anything on that file. Sorry for the
trouble.

yangsong8:
Cool, thanks.

yangsong8:
- [ ] This project does not include any db change, we do not need to merge db/schema.rb.

app/models/participant.rb
```diff
@@ -168,4 +168,8 @@ def get_scores(questions)
 
       scores
     end
+
+    def self.find_participant_byassign(user_id,assign_id)
```
yangsong8: You should also check if the participant.type is "AssignmentParticipant".

jli53: We rewrite this function according to the original code "participant =
Participant.where(user_id: session[:user].id, parent_id:
assignment_id).first".
We did not check if the original code miss some important checks. I think
for this part, our task is just  to improve format and style.

PR #578, opened 2015-08-30
currentpull

ssweene2:


ssweene2:
Cannot pull without a compare

PR #587, opened 2015-10-30
E1571. Refactoring AssignmentsController and AssignmentForm

ssharm21:
Refactor edit and update method.
Remove irrelevant comments from the create action.
associate_assignment_with_course and remove_assignment_from_course logic should be moved to model.
Change to new redirect method rather using controller and action explicitly.
Refactor the update_due_dates from AssignmentForm. using the DRY principle
Refactor code to follow ruby style guide and Rails4 syntax.


yangsong8:
I pulled and test the code. I found that metareview deadline cannot be saved:
image
image

image

After clicking "save", the page refreshes and the metareview deadline is not saved:
image

The metareview_deadline is not saved in DB, either.

yangsong8:
I tried to give a super long name to an assignment to test if Expertiza can return reasonable error message.

image

After clicking "save" I got error message below:
image

It seems the error message returning does not work as it supposed to.

yangsong8:
This team changed the name of a view (assignments/list->assignment/index). However I am not able to use this view and I suspect this view is no longer used since we have tree display.

image

yangsong8:
Just one more minor issue, it may not be a good idea to hard_code the lines like "DEADLINE_TYPE_METAREVIEW = 5", they are in dead_line_types table. And only hard wire from type 5 to type 8 is not a good idea.

ssharm21:
@yangsong8  by when the issues need to be fixed?

yangsong8:
Hi, I suggest you should focus on your exams for now. It would be really mean if I ask you guys to fix those issue in the final week. I would be a good news if you can fix those issues before Dec. 17th but *it is not a requirement*.

Good luck to your finals every one!

ssharm21:
thanks :+1:

PR #591, opened 2015-10-30
E1569: Refactored ReviewMappingController

rrpod:
Removed delete_rofreviewer
Removed delete_participant
Removed list_sortable
Modified response_report to use ActiveRecord instead of SQL
Moved add_user_to_assignment from ReviewMappingController to ParticipantsController
Method get_team_from_submission moved to AssignmentTeam model
Renamed delete_all_reviewers to delete_outstanding_reviewers
Removed delete_participant
Method release_reservation renamed as release_mapping.
Removed delete_review
Wrote Test Cases for ReviewMappingController


yangsong8:
When I merge this project, I did not see any place that "add_user_to_assignment" is used. Also, in this method there is one method call with missing arguments:
image

Another issue, this team moved  "get_team_from_submission" from review_mapping_controller to assinment_team.rb. However they left the method call and this method call is gonna fail for sure.
image

PR #593, opened 2015-10-31
E1554. Refactoring TeamsConrolller.rb

roshini123:


roshini123:
The database.yml.example was not proper. So we changed that and now will create a new pull request.

PR #595, opened 2015-10-31
E1556: Refactoring SuggestionController.rb

Hongyi23:
1. Change the code to be in accord with Rails 4 syntax.
2. Split Approve_suggestion().
3. Three functional tests.


app/controllers/suggestion_controller.rb
```diff
@@ -125,66 +156,32 @@ def approve_suggestion
...
-        ).deliver
       end
+    else
+      #if this team has topic, Expertiza will send an email (suggested_topic_approved_message) to this team
+      send_email
     end
-    redirect_to :action => 'show', :id => @suggestion
+  end
+
+  def approve_suggestion
```
yangsong8: There is an approve and an approve_suggestion method, is this sort of confusing? At least one of them should be private.

spec/controllers/suggestion_controller_spec.rb
```diff
@@ -0,0 +1,307 @@
...
+describe "SuggestionController" do
+
+  describe "#Test1:approve suggestion when having a topic in waitlist" do
+  
+    # Prepare for test
+    before(:each) do
+        @assignment_id = 711
+        @topic_id = 2864
+        @team_id = 27816
+        @assignment = Assignment.find_by_id(@assignment_id)
```
yangsong8: So you are assuming that assignment 711 is in the testing db. Unfortunately it is not true on Travis CI. Travis CI only runs "db:seed" before running tests.

app/controllers/suggestion_controller.rb
Winbobob: I have just changed the ```approve``` method to private.

PR #596, opened 2015-10-31
E1565. Refactoring admin_controller and course_controlle

jaithrik:


app/controllers/admin_controller.rb
```diff
@@ -93,12 +97,38 @@ def show_instructor
...
+      @role = Role.new(:id => nil, :name => '(none)')
+    end
+  end
 
   def remove_instructor
     User.find(params[:id]).destroy
     redirect_to :action => 'list_instructors'
   end
 
+  def remove_administrator
```
yangsong8: The method remove_administrator and remove_super_administrator look scary to me.

app/controllers/admin_controller.rb
```diff
@@ -40,6 +39,9 @@ def new_instructor
   #     end
   #   end
   # end
+ # def show
```
jaithrik: It's not significant.. I put that to try some redirects. It didn't work out so I just commented it. Will remove it now

app/controllers/admin_controller.rb
jaithrik: I've removed the remove super admins method. I put that method initially because it was given in our project description.

 The delete admin functionality was broken earlier. I tried to fix it by creating the remove_admin method. It is similar to the remove_instructor method.

PR #598, opened 2015-10-31
E1558. Refactoring ReviewResponseMap.rb 

adeeshag:
Some more files have been changed in the model due to the change in a method name and the models invoking the method name.


adeeshag:
Updated a few files

PR #600, opened 2015-10-31
E1551. Refactoring response_controller.rb

viswaraavi:


app/controllers/response_controller.rb
```diff
@@ -175,12 +76,13 @@ def edit
 
   #Update the response and answers when student "edit" existing response
   def update
+    return unless action_allowed?
```
yangsong8: this line is not necessary

yangsong8:
Merged by Dr. Gehringer.

PR #604, opened 2015-10-31
Assignment E1574

shrenujgandhi:
Feature test for assignment submission by student.
Uses : development database and requires prior execution of assignment creation file


yangsong8:
Hi team, thank you for your contribution. However I will not merge your pull request because your tests are written under assumption that "student13" is in the testing db. However it is not true. When the automated tests run, there is only an almost-empty db with full db schema and data defined in seeds.rb

PR #607, opened 2015-10-31
E1560: Refactor PopUpController and ParticipantsController

viveksubbarao:
Refectored code in the PopUpController.rb and ParticipantsController.rb


yangsong8:
popup_controller.team_users_popup is broken. 
In production:
image

In refactored code:
image

yangsong8:
The participant_controller.inherit does not work as it supposed to. I create an assignment "123" under CSC/ECE517 Spring 2015 (which has course participants). Then when I check the participant list in "123" assignment and it was empty.

image

image

yangsong8:
The method bequeath_all in participant_controller does not work as it supposed to be, either. 

I added myself as a participant to my newly created "123" assignment. I was not a course participant of "CSC/ECE514 Spring 2015" and I am a participant of this assignment in this course. If I click "copy participants to course", I should be added as a course participant of  "CSC/ECE514 Spring 2015".

image

Below is the error message I got.

image

PR #608, opened 2015-11-01
E1572: Feature Test for Assignment Creation via instructor

gmeneze:
http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Fall_2015/ossE1572VGA


db/schema.rb
```diff
@@ -698,24 +699,49 @@
...
     t.string  "persistence_token",         limit: 255
     t.string  "timezonepref",              limit: 255
     t.text    "public_key",                limit: 65535
-    t.boolean "copy_of_emails",            limit: 1,     default: false
+    t.boolean "copy_of_emails",                          default: false
   end
 
   add_index "users", ["role_id"], name: "fk_user_role_id", using: :btree
 
+  create_table "users_bkp", id: false, force: :cascade do |t|
```
yangsong8: What is this table supposed to be?

spec/features/assignment_creation_spec.rb
```diff
@@ -0,0 +1,645 @@
...
+         visit root_path
+         fill_in('login_name', :with => 'instructor6')
+         fill_in('login_password', :with => 'password')
+         click_on('SIGN IN')
+         expect(page).to have_content('Manage',:wait=>2)
+         within(".content") do
+         click_on("Assignments")
+         click_button 'New private assignment'
+         fill_in('assignment_form_assignment_name',:with => GenerateAssignmentName())
+         select('CSC 517 Fall 2010', from: 'assignment_form_assignment_course_id')
```
yangsong8: This is test is not going to pass Travis CI because in the test db there will not be such course. In travic CI, there is only data defined in seeds.rb. It is not a good idea to suppose the full production db is available for testing. However, it is okay that you create some courses,  assignments, or users for testing use.

PR #609, opened 2015-11-01
E1567. Refactoring User and UsersController

tarunchhabra26:
Added changes as per the refactoring requirements.


nikhiljosyabhatla:
Thank you for your contribution to the Expertiza project. We won't be merging your code since you could not make the travis build pass. Although, there were some good refactoring on your part, we would not want code that would break the Travis build to be added.

tarunchhabra26:
Hi Nikhil,

I had spoken to Yatish about the broken travis build and I was told that this wasn't because of our commits. Rather there was something wrong at the expertiza team's end. Would request you to please clarify this point once.

yangsong8:
I tried to create a new account "try2delete". When I tried to delete this user the view was not displayed correctly after I clicked "OK".

image

The refactoring on the get_user_list in user.rb looks not completed to me, too.

PR #610, opened 2015-11-01
E1563. Refactoring the Questionnaires Controller

agarg9:


yangsong8:
Thank you guys for your contribution. However we may not merge this pull request because this pull request tries to separate quiz questionnaires with other types of questionnaires (e.g. peer-review questionnaire, teammate review questionnaire). For other type of questionnaires, they do not have separate controllers, or even separate views. This is a problem with quiz questionnaires: the code dealing with quiz questionnaires is different from the code dealing with other types of questionnaires. The better design is to make use of polymorphism (those different types are all sub-classes of questionnaire class, and quiz questions can be sub-class of question class). We hope we can use the same code to deal with all types of questionnaire, not dealing with them separately. Thank you again for your effort.

PR #615, opened 2015-11-10
E1574. Feature Test for Assignment Submission

kunalb6:
Created Feature test for assignment submission by student using rspec and capybara.

During the course of our testing, we encountered two potential bugs in the Expertiza system. These are:
A student can upload file even after the assignment is finished.
A student can upload an .exe file to any assignment.

Detailed information can be viewed at http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Fall_2015/oss_E1574_BKS#Existing_Bugs.

### DIFF
--- a/spec/assignment_setup.rb
+++ b/spec/assignment_setup.rb
@@ -1,0 +1,58 @@
+# Method to create assignment before running feature test for assignment submission by student
+def create_assignment(name,due_date)
+
+  # Creates assignment
+  @assignment = Assignment.new(
+      :name => name,
+      :directory_path => "csc517/oss",
+      :instructor_id => 6,
+      :max_team_size => 1,
+      :course_id => 73,
+      :wiki_type_id => 1,
+      :reviews_visible_to_all => 0,
+      :staggered_deadline => 0,
+      :allow_suggestions => 0,
+      :review_assignment_strategy => "Auto-Selected",
+      :require_quiz => 0,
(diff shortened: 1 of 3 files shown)


spec/assignment_setup.rb
```diff
@@ -0,0 +1,58 @@
...
+  # Generate a node for the created assignment
+  node= Node.new(
+      :parent_id => 3920,
+      :node_object_id => @assignment.id,
+      :type => "AssignmentNode")
+
+  node.save
+
+  # Add "student13" as a participant to the created assignment
+  participant = Participant.new(
```
yangsong8: I see one issue here: you are assuming that this user_13 exists in the test db. There might be some potential problem. When Travis CI run the test, it will load all the db:migration first and then do rake db:seed. So b4 running the test, you can only assume that the data defined in seeds.rb is in the test db.

spec/features/student_submission_spec.rb
```diff
@@ -0,0 +1,101 @@
+ENV["RAILS_ENV"] ||= 'development'
+require 'rails_helper'
+require 'spec_helper'
+require 'assignment_setup'
+
+RSpec.feature 'assignment submission when student' do
+
+  active_assignment="FeatureTest"
+  expired_assignment="LibraryRailsApp"
```
yangsong8: Again, this assignment does not exist in seeds.rb.

PR #624, opened 2015-12-04
E1587: Improvements to password recovery and repeated login failures

rahulpshah:


app/controllers/invitations_controller.rb
```diff
@@ -0,0 +1,143 @@
+<<<<<<< HEAD
```
yangsong8: Why this file is changed? This does not seem to be relevant to this project.

db/migrate/20151120052354_add_loginattempts_nextlogin_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddLoginattemptsNextloginToUsers < ActiveRecord::Migration
+  def change
+  add_column :users, :login_attempts, :integer, default: 0
```
yangsong8: Can I try to login as other persons to block their accounts?

app/controllers/auth_controller.rb
```diff
@@ -1,39 +1,92 @@
...
+
+  def exponential_backoff(user)
+    # function to handle exponential backoff afterfailedlogin attempt
+    user.login_attempts=user.login_attempts+1
+    interval=2**(user.login_attempts-3)
+    user.next_login_time=DateTime.now+interval.minutes
+    user.save
+    logger.warn "Failed login attempt: Account Blocked"
+    flash[:error] = "Account is Blocked for #{interval} minutes"
+    render 'content_pages/relogin'
```
yangsong8: Is this possible that the user change the url from relogin to login after the 1st try fails?

PR #628, opened 2015-12-04
E1586: Anonymous chat between author and reviewer

nikunj91:
During reviews, many times reviewers will be having questions about the submitted material and if it's a blocker there is no way that the reviewer can progress further with the review.  As a result, reviewers will end up submitting the empty review forms. The aim of this project is to provide a ‘Ask for Quick Help’ button on the review forms such that reviewers can ask questions to authors.

Below features have been implemented:
Provide a ‘TextArea’ input field on review forms, where reviewers can type in their questions.
Create a new button ‘Ask for Quick Help’ on the review forms.
Upon click of this button, an email should be sent to the authors that questions were asked about their submitted work.
The email should have a link, using which the authors can answer the questions.
Reviewers should be notified by email, when authors answer the questions.


yangsong8:
Hi 1586 team, I know it is the busiest time of the semester, but do you want to make further changes on your project or you would rather we grade it as-is? Thanks.

app/controllers/response_controller.rb
```diff
@@ -330,6 +331,17 @@ def redirection
     end
   end
 
+
+  def submitted_question
```
yangsong8: I suggest you guys rename it to "initiate_chat" and put this method in review_chats_controller.rb if possible.

yangsong8:
Hi guys, I just met Nikhil Ch. It is okay if you can do the changes by 16th. I will give you more code review and comment on the places you may consider to change.

nikunj91:
Hi Yang, we would certainly like to update the current code and it would be great if the grading would happen after that. We will do the changes by 16th as mentioned by Nikhil. Let us know more areas where we can refactor the current implementation in addition to those we discussed during the demo

yangsong8:
Yes I will give you more code review later.

db/migrate/20151117005827_create_review_chats.rb
```diff
@@ -0,0 +1,13 @@
+class CreateReviewChats < ActiveRecord::Migration
+  def change
+    create_table :review_chats do |t|
+      t.integer :assignment_id
```
yangsong8: The biggest thing I would suggest you change is that for the this migration, you may remove the assignment_id, reviewer_id and team_id. You only need to add a response_map_id and this foreign key points to a record in response_map with type="ReviewResponseMap".
After that:
response_map = ReviewResponseMap.find(review_chat_object.response_map_id)
team_id = response_map.reviewee_id
reviewer_id = response_map .reviewer_id
assignment_id = AssignmentParticipant.find(reviewer_id).parent_id

app/controllers/review_chats_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    @review_chats = ReviewChat.all
+  end
+
+  # GET /review_chats/1
+  def show
+    @review_chat = ReviewChat.find(params[:id])
+    @assignment_id=@review_chat.assignment_id
+    @reviewer_id=@review_chat.reviewer_id
+    @team_id=@review_chat.team_id
+    @chat_log=ReviewChat.where(:reviewer_id => @reviewer_id).where(:team_id => @team_id)
```
yangsong8: After you change the migration, you can use 
@chat_log=ReviewChat.where(:response_map_id => @review_chat..response_map_id)

The idea that we can use a response_map_id to get the chatting thread.

nikunj91:
Sure, we will do the migration and once done will change the usage to include the response_map_id instead

app/views/response/response.html.erb
```diff
@@ -86,6 +87,21 @@ jQuery(document).ready(function() {
     <%= hidden_field_tag('return', @return) %>
     <%= hidden_field_tag  "isSubmit", :id=>"isSubmit" %>
 <% end %>
+
```
yangsong8: This part actually is not related to giving peer-review (responses), but to initiating review_chats. So this may not be a good place. I you add a link on the "others' work" page, next to "begin" "edit" or "update" of each response map. In this way the chatting feature would be more visible.

db/migrate/20151117005827_create_review_chats.rb
nikunj91: We have performed this migration and pushed the code to our git. We are making the other changes that you suggested us in the meanwhile

app/controllers/review_chats_controller.rb
nikunj91: We created a method get_chat_log in the model and have already implemented this and pushed to git

app/controllers/response_controller.rb
nikunj91: We have refactored the method to initiate_chat and pushed it

app/views/response/response.html.erb
nikunj91: We have updated the code so that the link is now shown in the 'others work' page next to the begin, update or edit

nikunj91:
Hi Yang,
We have incorporated all the changes that you have suggested and pushed the same as shown above.

PR #629, opened 2015-12-05
E1576 Refactoring submitted content (hyperlinks and files)

guru1989:


app/models/assignment_team.rb
```diff
@@ -89,8 +89,27 @@ def path
     self.participants.first.dir_path
```
yangsong8: There are 2 "path" defined, one in line 88 and another in 287.

yangsong8: One more thing, line 82, that hyperlinks method is not correct..

app/models/assignment_team.rb
```diff
@@ -265,6 +284,25 @@ def update_dirctory_num_for_new_member(new_participant)
     end
```
yangsong8: for line 273, this method should be removed!!

app/views/assignments/list_submissions.html.erb
```diff
@@ -55,13 +55,13 @@
  			
  			<% @participants.each do |participant| %>
```
yangsong8: Why @participants.each is still needed?

app/views/assignments/list_submissions.html.erb
```diff
@@ -55,13 +55,13 @@
...
     			<%= "Hyperlinks:<br/>".html_safe if !participant.hyperlinks.blank? and participant == @participants.first %>
-	    		<% participant.hyperlinks.each do |link| %>
+                <% participant.hyperlinks_array.each do |link| %>
 	    			<a href="<%= link %>" target="_blank"><%= link %></a><br/>
 	    		<% end %>
 	    	<% end %>
     		<!--Display submitted files-->
  			<% @participants.each do |participant| %>
```
yangsong8: Again why @participants.each is still needed? Cant we do @participants.first..submitted_files and save a loop?

db/migrate/20151128045237_remove_submitted_hyperlinks_directory_num_from_participants.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveSubmittedHyperlinksDirectoryNumFromParticipants < ActiveRecord::Migration
+  def change
+    remove_column :participants, :submitted_hyperlinks, :text
```
yangsong8: When you remove the column, you do not need to specify the type (text)

PR #640, opened 2015-12-05
E1581. Create integration tests for the student interface using capybara and rspec

xhy279:
E1581： we have eight tests which are corresponding to eight features in our design document.
add_member_spec.rb	
bring_up_review_form_spec.rb	
fill_out_review_forms_spec.rb	
review_feedback.rb	
select_a_submission_to_review_spec.rb	
select_a_topic_spec.rb	
student_sign_in_spec.rb
student_view_score.rb


spec/features/add_member_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+require_relative './helpers/login_helper'
+require 'selenium-webdriver'
+include LogInHelper
+
+
+feature 'Add someone to a team' do
+  scenario 'send the invitation as sender and receive it', :js=>true do
+
+    # sign in
+    log_in('student4346', 'password')
```
yangsong8: Hi, you need to take a look on this post: https://piazza.com/class/idkvxa78cn13c9?cid=403.

For this kind of test you cannot assume that "student4346" and assignment "Ethical analysis 2" are in the testing db. When the test starts, there is only the data in seeds.rb in the testing db. However, the "admin" account is defined and you can use admin to create students' account and assignment before you do your test.

xhy279: I understand that may not in accord with the testing convention; but the I
think the meaning for integration test is through simulating user behavior
to test the functions. Using development database can also do that.

yangsong8:
well, yes, those tests mean something when they run on production db. However those test will fail on Travis CI so your project (which is a testing project) will not be merged. It would be a pity that you guys work so much but this work finally got thrown away. If you have some time I do suggest you make some change. You do not need to change existing tests, only need to add some code to insert the necessary users and assignment.

Thanks and good luck to your finals.

xhy279:
Can you give us a quick example? And because my teammates and I are all
facing finals; so can we switch the demo after 11th. I think we really have
no time on this until 11th.  Thank you!

yangsong8:
You can take a look on this test: https://github.com/expertiza/expertiza/blob/master/spec/models/questionnaire_spec.rb.

xhy279:
Thanks！

xhy279:
We've updated our test cases to use data we provided in spec/features/db/TestData.sql. Our tests refresh the db every time before each test.

PR #642, opened 2015-12-05
E1582. Create integration tests for the instructor interface using capybara and rspec

junjieshen:


junjieshen:
E1583 Create integration tests for instructor interface using Rspec and Capybara

spec/features/instructor_interface_spec.rb
```diff
@@ -0,0 +1,168 @@
+require 'spec_helper'
+require 'rails_helper'
+require 'selenium-webdriver'
+
+describe "E1582. Create integration tests for the instructor interface using capybara and rspec" do
+
+  feature "Test1: Instructor login" do
+    scenario "with valid username and password" do
+      login_with 'instructor6', 'password'
```
yangsong8: Could you take a look on this post: https://piazza.com/class/idkvxa78cn13c9?cid=403.

This test is assuming that the "instructor6" is in the test db and this is not true on the testing db. You can only assume the testing db has the data defined in seeds.rb.

PR #654, opened 2016-03-21
E1604 Calibration Tests

EricHorton:
Testing the calibration feature.


spec/features/calibration_spec.rb
```diff
@@ -0,0 +1,667 @@
...
+    create :team_user, team: @team2, user: @student2
+    create :review_response_map, assignment: @assignment, reviewee: @team2
+
+
+
+  end
+
+  #creates a questionnaire, assigns it to the assignment, fills out the questionaire,
+  #displays the questionaire response, checks to make sure the score is there
+  it 'create a questionnaire, fill it out, display results', :js => true do
```
Winbobob: This test case is almost the same as [https://github.com/EricHorton/expertiza/blob/master/spec/features/calibration_spec.rb#L461](https://github.com/EricHorton/expertiza/blob/master/spec/features/calibration_spec.rb#L461). You had better to write a method to make the code DRY.

DevArenaCN:
Test cases code have been modified to make it more DRY

PR #656, opened 2016-03-22
E1600. Self-review feature

sujithktkm:
**A new migration was added to add a new columns to assignments table.
**Tests are yet to be fixed -> Tests were fixed in commits made during later part of submission.


app/models/user.rb
```diff
@@ -165,8 +165,9 @@ def email_welcome
   end
 
   def valid_password?(password)
-    Authlogic::CryptoProviders::Sha1.stretches = 1
-    Authlogic::CryptoProviders::Sha1.matches?(crypted_password, *[self.password_salt.to_s + password])
+    # Authlogic::CryptoProviders::Sha1.stretches = 1
+    # Authlogic::CryptoProviders::Sha1.matches?(crypted_password, *[self.password_salt.to_s + password])
+    true
```
yangsong8: Please change this back.

sujithktkm: Changed in 5092d0f9ba2980a7354593616cf065ca851e2716.

app/controllers/response_controller.rb
```diff
@@ -211,13 +217,21 @@ def create
     redirect_to :controller => 'response', :action => 'saving', :id => @map.map_id, :return => params[:return], :msg => msg, :error_msg => error_msg, :save_options => params[:save_options]
   end
 
+  #E1600
+  #Added paramps[:return] value for 'SelfReviewResponseMap'
   def saving
     @map = ResponseMap.find(params[:id])
+    if(@map.type == "SelfReviewResponseMap")
```
yangsong8: Is there any way that you can make params[:return] equals "selfreview" from view? Making this change in the controller does not seem to be a good idea.

app/models/review_response_map.rb
```diff
@@ -182,4 +182,13 @@ def self.final_versions_from_reviewer(reviewer_id)
     end
     review_final_versions
   end
+
+  def self_review_response_map?
```
yangsong8: Is this method used?

saiteja09: We are not using that method. Wrote it when we started project but forgot to remove it. I will take care of this method.

app/controllers/response_controller.rb
saiteja09: I will check on this and get back to you soon.

saiteja09: I checked if I could make a change to accommodate this return value in view, but I think it would be too messy as the response view is already complicated and making changes there would make reading the code more difficult. I think it is better to have it in controller itself for maintenance purposes

app/models/review_response_map.rb
sujithktkm: This method is removed in eb79aa523fc4f38510bbc68ac0dd474840bbcf47

yangsong8:
Partially merged with minor fix. Thank you for your work guys!

PR #661, opened 2016-03-22
E1613.  Refactor sign_up_sheet_controller.rb and sign_up_topic.rb

rgkavodkar:


app/assets/javascripts/tree_display.jsx
```diff
@@ -215,16 +215,20 @@ jQuery(document).ready(function() {
               </span>
             )
             if (this.props.has_topic) {
```
yangsong8: This empty if can be removed, right?

spec/features/instructor_interface_spec.rb
```diff
@@ -52,6 +52,17 @@
     end
   end
 
+  describe "View Publishing Rights" do
```
yangsong8: Hi everyone, please refer to point 5 under "On all refactoring" in the OSS project description. Login_as instructor 6 will fail since in test db there is no such user.

app/assets/javascripts/tree_display.jsx
rgkavodkar: Yes, this can be removed

PR #665, opened 2016-03-28
E1611 refactoring team.rb, assignment_team.rb, course_team.rb

nikraina:


nikraina:
need to change database.yml and then pull

PR #666, opened 2016-03-28
E1611 refactoring team.rb, assignment_team.rb, course_team.rb(second attempt)

nikraina:
second attempt after reverting the changes done to database.yml on local system


nikraina:
Failed because  .travis.yml had::
- cp config/database.yml.example config/database.yml.example
instead of 
 - cp config/database.yml.example config/database.yml
made the change now requesting a new pull request

app/models/assignment_team.rb
```diff
@@ -42,43 +62,52 @@ def topic
...
   def has_submissions?
-    list_of_users = participants;
-    list_of_users.each { |participant| return true if participant.has_submissions? }
-    false
+    (self.submitted_files.length > 0) or self.submitted_hyperlinks.blank?
   end
 
+  #Response Map for each reviewer/reviewee
   def reviewed_contributor?(contributor)
+    #ReviewResponseMap.all showing error but somehow its working, incase it fails please uncomment the following line
```
yangsong8: I cannot find any place that this method is called? Could you delete this if you can confirm that is it not called anywhere?

nikraina: Yes, you are right I could not find any usage as well hence i have deleted the method.

Winbobob:
After merging your project, I found that the format of export has something wrong. The last column should be a team member, but your format is assignment name. What's more, actually you store team type instead of assignment name for that column.

nikraina:
Hi Zhewei,

We have not changed the format of excel.
The behaviour is same as it was earlier. If you upload the downloaded excel
it works fine.
While doing refactoring we had similar doubts but whne we checked the
functionality worked correctly and the behaviour remained same after
refactoring we did not attempt to change it.

If you have more doubts please let me know. We can meet up and discuss this
as well if you want so.

Winbobob:
Actually, you did a good job! Since the format of CSV did not mention in OSS writeup, we will not deduct points for this part.

Thanks,
Zhewei

PR #690, opened 2016-04-30
E1624 Compose functional tests for Suggest topics of student functionality (3rd version)

hawkwoodye:
Notice that we found the students of 2064,2065,2066 will randomly create one of them during each describe in the test, so we added two fix students: studenta and studentb in the spec/factories/factories.rb file.


hawkwoodye:
We removed studenta and studentb in factories.rb, and use create_list participant, 3 instead. Also fixed the expect Title meaningless lines.

yangsong8:
I did not see studenta and studentb in spec/factories/factories.rb file as announced earlier? Have that been changed?

hawkwoodye:
Dear Yang,

I've removed them and use the original factories.rb file this morning as
Zhewei's asked last night.
So you won't see any studenta or studentb in the latest version.

Thanks,
Ye

PR #751, opened 2016-10-25
E1655: Improve Email Notifications 

cbaruah:
E1655 : Improve Email Notifications


yangsong8:
I checked you pull request.

- Most of the functions work.
- But your design does not follow the way that Expertiza works. You added a method 'send_sync_message' to MailerHelp and make it call the sync_message in Mailer class. This adds confusion because some code may wrap the information of email in 'defn' and call 'send_sync_message', and some caller may call 'MailerHelper.send_sync_message(***).deliver'.

app/helpers/mailer_helper.rb
```diff
@@ -11,4 +11,18 @@ def self.send_mail_to_user(user, subject, partial_name, password)
       }
     })
   end
+
+  def self.send_sync_message(user, subject, partial_name, type, obj_name)
```
yangsong8: I'm not so convinced that this method should be in helper.

PR #760, opened 2016-10-28
E1654. Improve date-picker and deadlines

DJaySathe:
What it does: allow instructor to define deadlines such as the assignment, peer review. 
What’s wrong with it: 
The current date picker doesn't let you change dates easily. Unless you spell the deadline out to the minute, it changes the deadline to the current time! 
When doing testing, we very often want to test an assignment in, say, the submission or review phase, even after the assignment is complete. In order to do this, we need to change several due dates from the UI. It would be a lot easier if we just had a command that said, "set all due dates ahead by x days.
What needs to be done:
Replace the date time picker in _due_dates.html.erb
Find a more user friendly JQuery datetime picker that support time zones and allow users to define deadline by date and time (hour).
Allow the user to change the timezone, but try guess the appropriate timezone for each entry. E.g., set the default time zone to the same timezone as the owner/instructor of the course.
Allow users to set all due dates ahead by x days, and the algorithm calculates the exact date and time


yangsong8:
I believe this is duplicated.

PR #761, opened 2016-10-28
E1643. Integration test suggestion_controller

anubhabMajumdar:
I have created an integration test script for suggestion_controller.rb named topic_suggestion2_spec.rb. It describes 6 scenarios and achieves 90% code coverage of suggestion_controller.rb.


Winbobob:
- Dirty pull request
 - Tests are similar to existing one

anubhabMajumdar:
I think the tests are different from the existing ones in some aspects, especially in terms of flow. Also, the existing tests covered around 70-75% of the suggestion controller. Our code covers 90% of the suggestion controller. Plus the suggestion controller is a small class, and there weren't many scenarios left to handle after topic_suggestion_spec.

PR #772, opened 2016-10-28
E1666. Test team functionality

hli36:
E1666. Test team functionality
Modify rails_helper.rb. 
Add a file team_functionality_spec.rb.
Some strange errors come when we tried to modify our codes. Our codes works very well in our own environment. And even if we pull it from GitHub, it works too. But it can't pass the  Travis CI check. So we revert our version to c64d6f0, which passed the check. Please ignore the commits after c64d6f0.
You can run team_functionality_spec.rb and see 6 tests passed.
Actually we write our codes together in hunt library. Although you can only see the commits by me, Abhinav and Fuxing also contributes to the test codes for different scenarios.

### DIFF
--- a/spec/features/team_functionality_spec.rb
+++ b/spec/features/team_functionality_spec.rb
@@ -878,7 +878,7 @@ describe "team functionality testing case 5", type: :feature do
 end
 
 
-#case 6 2 student selet the same topic should be in different teams
+#case 6 2 student select the same topic should be in different teams
 describe "team functionality testing case 6", type: :feature do
   before(:each) do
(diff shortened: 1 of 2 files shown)


spec/features/team_functionality_spec.rb
```diff
@@ -0,0 +1,1036 @@
...
+    expect(page).to have_content "Edit name "
+    expect(page).to have_content "student2064"
+    expect(page).to have_content "student2065"
+    expect(page).to have_content "Leave team"
+
+
+
+
+    end
+
```
yangsong8: Why so many white spaces?

Issue #669, opened 2016-04-05
Copy assignment teams into course or course teams to assignemt


yangsong8:
We have methods related to those 2 functionalities, but there is no way to test this feature from UI.

We should also see the code done in E1611.

yangsong8:
We looked at this issue again. We can find the "inherit teams from course" in "create team icon"->"create team" -> "Inherit teams from course". Not sure if this work (have to create a new assignment to test).

yangsong8:
But two further issues:
If an assignment does not belong to any course, "Inherit teams from course" should not be an option.
If one instructor click "craete team icon" on a course node -> "create team", the UI still displays "Inherit teams from course". This is not correct. Instead, it should offer the option for copying assignment teams to course.

yangsong8:
- Inheriting teams from course to assignment has been fixed.
- If an assignment does not belong to any course, the option of showing "inheriting teams from course" will not be displayed.
- "Inheriting teams from course" will not be displayed if the user is editing course teams.

Issue #331, opened 2013-08-21
Bad UI for copying teams from course to assignment (and also from assignment to course)


efg:
Teams can be copied from an assignment to a course, or vice versa.  But it's not very intuitive how to do this.  To copy a team from a course to an assignment, you need to go to the assignment, then choose "Create teams" from the popup menu, then click "Create team" down at the bottom of the page, then click the "Inherit teams from course" button at the bottom of the page.  It would be so much better if "Inherit teams from course" appeared on the "Create teams" page (one page less to click through").

When copying teams from an assignment to a course, one must copy them one by one, by clicking on the Copy icon next to each team.  There should be a way to copy _all_ teams from an assignment to a course.

Issue #303, opened 2013-06-18
Error trying to copy team to course when there is no course 


efg:
When someone selects "Copy team to course" from the Create_teams screen for a particular assignment, and the assignment is not part of a course, we get a NoMethodError in TeamController#bequeath.

To fix this, the "Copy team to course" icon should not appear on the Create_teams screen when the assignment is not part of a course.

yangsong8:
This has been done by E1651.

PR #777, opened 2016-10-28
E1651. Course Teams and submission directories

jitinnagpal1989:
Expertiza code v 1.0. Additional resources: http://wiki.expertiza.ncsu.edu/index.php?title=CSC/ECE_517_Fall_2016/oss_E1651


app/models/assignment_team.rb
```diff
@@ -16,6 +16,10 @@ def parent_model
     "Assignment"
   end
 
+  def self.parent_model_for_id (id)
```
yangsong8: This is not necessary, please check this line on very top of AssignmentTeam model:
`belongs_to :assignment, class_name: 'Assignment', foreign_key: 'parent_id'`

app/models/team.rb
```diff
@@ -247,18 +251,11 @@ def self.export(csv, parent_id, options, teamtype)
...
-      curr_assignment = Assignment.find(id)
-      team_name = Team.generate_team_name(curr_assignment.name)
-      team = AssignmentTeam.create(name: team_name, parent_id: id)
-      TeamNode.create(parent_id: id, node_object_id: team.id)
-    end
+  def self.create_team_and_node(id)
+    current_task = parent_model_for_id id
+    team_name = Team.generate_team_name(current_task.name)
+    team = self.create(name: team_name, parent_id: id)
+    TeamNode.create(parent_id: id, node_object_id: team.id)
```
yangsong8: The code here is tricky...
The `parent_id` will be the id of the 'parent_model' and the node_object_id will be the id of newly created team. 
Comment should be added to help future developers understand this part.

app/views/teams/list.html.erb
```diff
@@ -13,6 +13,15 @@
 <% else %>
   <% modelType = 'CourseTeam' %>
 <% end %>
+
+<% if modelType == 'AssignmentTeam' %>
```
yangsong8: This is not the write thing to do. 
On the teams/list page, if the current object is an assignment, you will see a section caller "inherit teams from course". But you will not see this section if the current object is a course.

app/models/team.rb
```diff
@@ -194,7 +194,11 @@ def self.import(row, id, options, teamtype)
     end
 
     if name
-      team = Team.create_team_and_node(id, teamtype)
+      if teamtype.is_a?(CourseTeam)
+        team = CourseTeam.create_team_and_node(id)
```
yangsong8: W8, there is no 'create_team_and_node' defined separately in AssignmentTeam model and CourseTeam model. What's the point of this change, then?

yangsong8:
I checked your pull request:

- You guys really spam the commits.
- Issue #669 got fixed in a different way. But I would not blame you because this is part of another OSS team and their work has been merged. You can see my latest comment on this issue.
- Issue #331 mentions that there should be a way to copy all the assignment teams to course (instead of clicking the copy icons one by one). I don't find this part has been implemented.
- Issue #303: your fix is correct.
- Minor coding issues, see in line comment.

PR #778, opened 2016-10-28
E1631. Refactor bidding interface

lzheng5:
Somebody tried to refactor the `sign_up_sheet_controller.rb` in the `master` branch. They modified some code but not `intelligent_topic_selection` view. Our original change has already been merged into the master branch, it was working.  

But if we tried to merge the code from `master`, the refactored code would fail. So we will wait for the refactoring to complete, and after that the build should automatically pass.


yangsong8:
@TheAkbar has merged and modified.

PR #781, opened 2016-10-28
E1652 Fix teammate advertisements and requests to join a team

sndesai92:
All commits have been done from a single machine using id sndesai92 but all team members have worked on the project.

Team members:
Saloni Desai
Shriyansh Yadav
Vaibhav Shinde


app/controllers/join_team_requests_controller.rb
```diff
@@ -51,13 +51,18 @@ def create
...
         @join_team_request.participant_id = participant.id
-        respond_to do |format|
-          if @join_team_request.save
-            format.html { redirect_to(@join_team_request, notice: 'JoinTeamRequest was successfully created.') }
-            format.xml  { render xml: @join_team_request, status: :created, location: @join_team_request }
-          else
-            format.html { render action: "new" }
-            format.xml  { render xml: @join_team_request.errors, status: :unprocessable_entity }
+        #need to check if this user has already sent a request
+        if !@join_team_request.firstRequest?
```
yangsong8: it would be better if this check is done before @join_team_request is created in line 47

app/models/response_map.rb
```diff
@@ -88,4 +88,8 @@ def self.delete_mappings(mappings, force = nil)
     end
     failedCount
   end
+
+  def self.isReviewed?(team_id,assignment_id)
+    review = ResponseMap.where(reviewee_id: team_id,reviewed_object_id: assignment_id).first
```
yangsong8: What is the returned value of this method? either true or false?

shindevaibhav: It will return nil if there is no record. otherwise it will return value of review(non-nil)

app/controllers/join_team_requests_controller.rb
shindevaibhav: We are doing it before saving it. But you are right.

yangsong8:
I have tested your pull request:

- Issue 421 -> does not exist anymore, confirmed.
- Issue 312 -> user who have requested to join a team once is till able to click "request invitation", would be better to allow user to change join team request information.

app/views/student_teams/view.html.erb
```diff
@@ -120,7 +120,7 @@
          <%= text_field 'team', 'name' %>
          <%= submit_tag "Name team" %>
        <% end %>
-     <% elsif !@student.team.full?%>
+     <% elsif !@student.team.full? && @student.assignment.has_topics? && @student.team.topic%>
```
yangsong8: Don't think the logic here is correct, should be 

(!@student.team.full? && !@student.assignment.has_topics?) || (!@student.team.full? && @student.assignment.has_topics? && @student.team.topic)

PR #786, opened 2016-10-28
E1656. Improve imports

olganam:
This is pull request for: E1656. Improve imports
Project explanation:  http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Fall_2016/oss_E1656


app/controllers/import_file_controller.rb
```diff
@@ -31,13 +74,14 @@ def importFile(session, params)
     delimiter = get_delimiter(params)
     file = params['file'].tempfile
 
+    logger.debug "the value of file is passed: #{file}"
```
yangsong8: This should be removed

yangsong8:
I tested your pull request,

- Export participants of assignment 754 with choice "All Paritcipants" and I got a list of students with "All Participants" as "team_id" and "no topic chosen" and topic. This does not look correct. This is wikipedia assignment and most students should have topics.

app/controllers/import_file_controller.rb
```diff
@@ -11,6 +11,49 @@ def start
...
+        when "ReviewResponseMap"
+          @array_expected_values = parse_line("Contributor, Reviewer1, Reviewer2,...,ReviewerN",',',params)
+        when "MetaeviewResponseMap"
+          @array_expected_values = parse_line("Contributor, Reviewer, Metareviewer1,Metareviewer2,...,MetareviewerN",',',params)
+        when "SignUpTopic"
+          @array_expected_values = parse_line("Topic identifier, Topic name, Max choosers, Topic Category",',',params)
+        when "SignUpSheet"
+          @array_expected_values = parse_line("Topic identifier, user1, user2,...",',',params)
+        else
+          @array_expected_values = parse_line("Team Name - optional, Team Member1,Team Member2, ... ,Team MemberN",',',params)
```
yangsong8: What type will fall into this else branch? "AssignmentParticipant", "CourseParticipant" and "CourseTeam" are not covered above, but I don't think they all share the same expected fields.

app/controllers/import_file_controller.rb
olganam: Will update. thanks

app/controllers/import_file_controller.rb
olganam: Will update. thanks

Rishabh2693:
Will Update the required change in all participants

PR #791, opened 2016-10-28
E1674. Refactor leaderboard.rb and write unit tests

keemen90:


Winbobob:
- Your refactor LGTM
 - Your Rspec test use too many `create`, which means too many DB connection. For unit tests, we prefer `build` and `double`.
 - As professor decided, we are not going to maintain this functionality any more, so I will close this pull request.

PR #792, opened 2016-10-28
E1670. Unit tests for answers.rb

anudeepvrm:


spec/models/answer_spec.rb
```diff
@@ -0,0 +1,165 @@
+require 'rails_helper'
+
+describe Answer do
+  let(:questionnaire) {create(:questionnaire,:id => 1)}
```
Winbobob: It is overkilled to use `create` in unit tests, using `double` or `build` instead.

spec/models/answer_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+  let!(:answer) { create(:answer,:question => question1,:response_id => 1)}
+  
+  describe "# test dependancy between question.rb and answer.rb" 
+    it { should belong_to(:question) }
+  
+  describe "#test get total score" do
+   				
+    it "returns total score when required conditions are met" do
+      # stub for ScoreView.find_by_sql to revent prevent unit testing sql db queries
+      allow(ScoreView).to receive(:find_by_sql).and_return([double("scoreview",weighted_score: 20,sum_of_weights: 5,q1_max_question_score: 4)])
```
Winbobob: Duplicated code, it will be better to extract a method to deal with.

anudeepvrm: do I make these changes now or should I wait for the build to get stable ? It is currently failing

spec/models/answer_spec.rb
anudeepvrm: Yeah I have used stubs and doubles everywhere except for testing sql queries. I know it is a bad practice to create test record and test queries but the prof told me to write general tests to see if db schema is up to date. Hence I had to use create for those sql tests.

Winbobob: OK

PR #799, opened 2016-10-29
E1653. Fix and improve rubric criteria

navisingh14:
1. Access control is given only to the owner of a questionnaire or his/her TA, to edit, delete, or toggle access
1.a Hard-coded ids have been replaced with generic codes
1.b User is shown the reason why he's not allowed to perform a particular action (customized comments) 

2. You can now import/export questionnaires from/to a CSV file
2.a The name of the exported questionnaire CSV file is the same as the questionnaire being exported 

3. Feature tests are done to check the working functionality of import/export and advice section


yangsong8:
I have checked your pull request:

- Issue 696 -> your logic is correct, but hard coded TA role id is not cool. In addition, if an instructor click the edit icon on a questionnaire which is not owned by him/her, the error message is:
image
This error message does not explain why the access is blocked.

app/controllers/questionnaires_controller.rb
```diff
@@ -572,20 +597,27 @@ def question_params
...
-
-    @questionnaire.questions << QuestionnaireHelper.get_questions_from_csv(@questionnaire, file)
+    questionnaire_id = (params[:id])
+    begin
+      uploaded_io = params[:csv]
+      File.open(Rails.root.join('spec', 'features/import_export_csv_oss/', uploaded_io.original_filename), 'wb') do |file|
+        file.write(uploaded_io.read)
+      end
+        file_data = File.read(Rails.root.join('spec/features/import_export_csv_oss/'+uploaded_io.original_filename))
+        a = QuestionnaireHelper.get_questions_from_csv(file_data,params[:id])
```
yangsong8: Why variable "a" is needed?

app/views/questionnaires/edit.xls.erb
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
```
yangsong8: What this file is for?

navisingh14: We were planning to use this in modifying the error message. Can we push the changes after removing this?

app/controllers/questionnaires_controller.rb
navisingh14: We were planning to use this in modifying the error message. Can we push the changes after removing this?

PR #800, opened 2016-10-29
 E1672. Remove duplicated code in feature tests

svenka15:
svenka15
nnatara2
vchitto


spec/rails_helper.rb
```diff
@@ -65,6 +82,60 @@ def login_as(user_name)
     click_button 'SIGN IN'
     stub_current_user(user, user.role.name, user.role)
   end
+  def instructor_login
+    describe "Instructor login" do
```
Winbobob: Why do you put some test cases into `spec/rails_helper.rb`?

Winbobob:
You had better extract duplicated lines of code into a new method instead of extracting the whole test cases.

PR #802, opened 2016-10-29
E1671. Unit tests for participants.rb hierarchy (assignment_participant, course_participant)

hsy12321:


spec/factories/factories.rb
```diff
@@ -324,4 +324,44 @@
...
+    course { Course.first || association(:course) }
+    can_submit true
+    can_review true
+    assignment { Assignment.first || association(:assignment) }
+    association :user, factory: :student
+    submitted_at nil
+    permission_granted nil
+    penalty_accumulated 0
+    grade nil
+    type "AssignmentParticipant"
```
Winbobob: The type of course_participant should be `CourseParticipant`.

spec/models/participant_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+
+
+describe Participant do
+
+  before(:each) do
+    @student = create(:student, name: "Student", fullname: "Student Test", email: "student1@gmail.com" , password: "123456", password_confirmation: "123456")
```
Winbobob: It is overkill to use `create` in unit tests.It would be better to use `double` or `build`.

Winbobob:
See #819

PR #803, opened 2016-10-29
E1644. Refactor and test TeamsController.rb

ajaypendyala:
Teammates:
Ajay Chandra Pendyala (apendya)
Mohit Sardar (msardar)
Bhaskar Sinha (bsinha)
The issues we worked on:
1. In delete method in teams_controller.rb, the variable “@signUps” should be changed to snake case. Please change it and make sure the delete method works for deleting both assignment teams and course teams. Write a test for this method.
2. Inheriting teams from course to assignment functionality can be found by clicking “Create Teams” icon then “Create teams”. But the “Inherit Teams From Course” option should not display when either 1) we are editing a course object or 2) the current assignment object does not have a course associated with it. Please fix the view for this, also write test for the inherit-team functionality.
3. The method create is called when an instructor tries to create a team manually. This should work for both creating assignment teams and course teams. Please make sure this functionality works and write a test for it. 
4. Write a test for the list method. This method lists all the team nodes, and you can expand each team node to see the user nodes as well. Please note that each team object has a corresponding team node object for use in the UI.
5. Write a test for the delete method. This method should work for both assignment teams and course teams, should work for either empty teams or teams holding a topic.


app/views/teams/new.html.erb
```diff
@@ -28,14 +28,15 @@
 <% end %>
 
 <BR/>
-<H2> Inherit Teams From Course </H2>
-<p> Use the teams that are currently defined for the course.</p>
-<%= form_tag :action => 'inherit' do %>
-  <%= hidden_field_tag 'id', @parent.id %>
-
-  <%= submit_tag "Inherit" %>
+<% if not @parent.instance_of?(Course) and not @parent.course.nil? %>
```
yangsong8: How about 
if @parent.instance_of?(Assignment) and n@parent.course
?

yangsong8:
I tried to test this pull request manually:

- manually created a team (works)
- manually added one team member to this team (didn't show the new team member name on teams/list)

ajaypendyala:
Thanks @yangsong8 for the detailed explanation and for merging our pull req.

PR #807, opened 2016-10-29
E1665. Test staggered-deadline functionality

popoosl:
Initial work, need improved in next days


Winbobob:
- You did not sync the remote repo!

Winbobob:
In scanario1, student2064 in review stage could review others' work.
However, student2065 is still in submission stage.
So actually, student2064 cannot review anything.
The reason is that the `review_allowed_id` of default submission deadline is `OK`, should be `NO`.
But it is not your fault.

PR #812, opened 2016-10-29
E1668.Test e-mailing functionality

RoyXue812:


spec/features/edit_emails_spec.rb
```diff
@@ -0,0 +1,95 @@
+require 'rails_helper'
```
Winbobob: You removed others duplicated code in `spec/features/delayed_mailer_spec.rb`, and created lots of duplicated code here.

PR #816, opened 2016-10-29
E1648 Add past-due assignments to task list

SaiSameer:


app/controllers/student_task_controller.rb
```diff
@@ -13,10 +13,27 @@ def list
...
     @students_teamed_with = StudentTask.teamed_students current_user
+
+    ######## Next Immediate Due Date ###############
+    @next_deadline = self.next_dead_line
+  end
+
+  def next_dead_line
+    ######## Next Immediate Due Date ###############
+    next_deadline = nil
+    @student_tasks = StudentTask.from_user current_user
```
Winbobob: `@student_tasks` has already been assigned in `list` method, you can reuse it by passing into `next_deadline`.

app/views/student_task/list.html.erb
```diff
@@ -46,6 +46,22 @@
     </span><br/>
   <% end %>
 
+    <strong>&nbsp;&nbsp;<span class="tasknum">&nbsp;<%= @latetasks.size.to_s %>&nbsp;</span> Late Tasks<br></strong><br>
+     <% @latetasks.each do |student_task|
```
Winbobob: Write the loop in view like this is really weird, had better put it in `student_task_helper.rb`.

db/migrate/137_add_finished_deadline_type.rb
```diff
@@ -0,0 +1,3 @@
+class AddFinishedDeadlineType < ActiveRecord::Migration
+  execute "INSERT INTO `deadline_types` VALUES (12,'Finished');"
```
Winbobob: Deadline_type `Finished` is not stored in DB originally. Actually, I do not think `Finished`  is a deadline type. It is just a stage. When mentioning deadline type, we are talking about submission deadline, review deadline, sth like this.

app/models/due_date.rb
```diff
@@ -83,4 +83,13 @@ def self.get_next_due_date(assignment_id, topic_id = nil)
     end
     next_due_date
   end
+  
+  #Find the due date at which the assignment was Finished.
+  def self.get_past_due_date(assignment_id, topic_id = nil)
```
Winbobob: This method looks not correct. You are looking at the first `AssignmentDueDate` and assign it to `Finish`?

Winbobob:
I think I will partially merge your project.

Winbobob:
See 393afaea412d3bd0c2b32ef02807ed7e35eed1b1

PR #819, opened 2016-10-29
E1671. Unit tests for participants.rb hierarchy (assignment_participant, course_participant)

kaps1489:


db/schema.rb
```diff
@@ -286,7 +286,7 @@
   add_index "menu_items", ["controller_action_id"], name: "fk_menu_item_controller_action_id", using: :btree
   add_index "menu_items", ["parent_id"], name: "fk_menu_item_parent_id", using: :btree
 
-  create_table "nodes", force: :cascade do |t|
+  create_table "nodes", force: :cascade do |t|1
```
yangsong8: Looks like this is a mistake.

kaps1489: oh yes i guess, this has been added accidentally while reviewing  the code

Gemfile
```diff
@@ -75,6 +75,7 @@ group :test do
   gem 'gherkin'
   gem 'guard-rails'
   gem 'guard-rspec'
+  gem 'factory_girl'
```
Winbobob: We already have `factory_girl_rails` in Gemfile, so I think you do not need to add a new gem here.

spec/models/participant_spec.rb
```diff
@@ -0,0 +1,50 @@
+
+require 'rails_helper'
+
+
+describe Participant do
+
+  before(:each) do
+    @student = create(:student, name: "Student", fullname: "Student Test", email: "student1@gmail.com" , password: "123456", password_confirmation: "123456")
```
Winbobob: It is overkill to use `create` in unit tests.It would be better to use `double` or `build`.

spec/models/course_participant_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+  describe "validations" do
+    before(:each) do
+      @course_participant = build(:course_participant)
+    end
+  end
+
+  describe "#copy" do
+
+    it "create a copy of participant" do
+      assignment = build(:assignment)
```
Winbobob: Duplicate code, it will be better to put them in `before(:each)` block

PR #827, opened 2016-10-31
Introduce a student view for instructors and goto future date

chaitanya91k:
Changes include a new textbox and button on the header for instructor for viewing as a student. After the instructor selects a student, he can go to a specific date and view student view for that day.


yangsong8:
Is this an OSS project? could you provide the project number or explain a little bit more on what it does?

chaitanya91k:
Hi,

Yes the changes are for an OSS project. Below is the wiki page for more
details.
http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Fall_2016/E1657._Introduce_a_Student_View_for_instructors

yangsong8:
I tested your pull request:
Here is what I got by impersonating user 'efg' (which was instructor6 in scrubbed db):
image
And here is what I got by logging in as admin: 
image

chaitanya91k:
* I do not know the reason why the date picker is not working, but found out that the date picker is browser based. It is working on Chrome but not on Firefox.
* Also about the "Switch to Student View" will be each student specific. Because a single instructor may take multiple courses. We do not know which view to show up. Thus we have decided to go with only student specific view.
* For testing 3 year old date, can you give us the database? As what ever case we tested with , the status appeared correct.
*The reason instructor needs to type in the studentID again is to give the instructor to be able to view as a different Student without adding new text bar. There are multiple texts already in the menu.

PR #828, opened 2016-11-01
E1632: Refactor delayed mailer.rb and scheduled task.rb

sjha4:
Refactor delayed job.rb and scheduled task.rb

### DIFF
--- a/app/models/drop_one_member_topics.rb
+++ b/app/models/drop_one_member_topics.rb
@@ -1,0 +1,19 @@
+class drop_one_member_topics < DeadlineType
+
+email_list(assignment_id)
+email=[]	
+assignment = Assignment.find(self.assignment_id)
+       
+if (assignment.team_assignment?)
+teams = TeamsUser.all.group(:team_id).count(:team_id)
+    for team_id in teams.keys
+      if teams[team_id] == 1
+        topic_to_drop = SignedUpTeam.where(team_id: team_id).first
+        topic_to_drop.delete if topic_to_drop #check if the one-person-team has signed up a topic
+       end
+     end
+   end
+  end
+email[]
+end
+end
--- a/app/models/drop_one_member_topics.rb
+++ b/app/models/drop_one_member_topics.rb
@@ -1,19 +1,20 @@
-class drop_one_member_topics < DeadlineType
+class DropOneMemberTopics < DeadlineType
 
-email_list(assignment_id)
-email=[]	
-assignment = Assignment.find(self.assignment_id)
-       
-if (assignment.team_assignment?)
-teams = TeamsUser.all.group(:team_id).count(:team_id)
+def email_list(assignment_id)
+  assignment = Assignment.find(assignment_id)
+  drop_one_member_topics if (assignment.team_assignment?)
+  emails =[]
+  emails     
+end
+
+def drop_one_member_topics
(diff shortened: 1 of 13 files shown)


app/models/drop_one_member_topics.rb
```diff
@@ -0,0 +1,20 @@
+class DropOneMemberTopics < DeadlineType
```
yangsong8: How to set this type of deadline up from UI?

app/models/drop_one_member_topics.rb
```diff
@@ -0,0 +1,20 @@
...
+
+def email_list(assignment_id)
+  assignment = Assignment.find(assignment_id)
+  drop_one_member_topics if (assignment.team_assignment?)
+  emails =[]
+  emails     
+end
+
+def drop_one_member_topics
+ teams = TeamsUser.all.group(:team_id).count(:team_id)
```
yangsong8: This is not correct. All the teams with only 1 member will be dropped, which is definitely not correct. There are many assignments in Expertiza require individual submissions. Instead, you should find the teams _in this assignment_ which have only one member.

app/models/drop_outstanding_reviews.rb
```diff
@@ -0,0 +1,21 @@
+class DropOutstandingReviews  < DeadlineType
```
yangsong8: Again, I don't know if we got this type of deadline supported from UI

app/models/drop_outstanding_reviews.rb
```diff
@@ -0,0 +1,21 @@
...
+
+	def email_list(assignment_id)
```
yangsong8: White space should be fixed.

app/models/drop_outstanding_reviews.rb
```diff
@@ -0,0 +1,21 @@
...
+    emails = drop_outstanding_reviews(assignment_id)
+    emails
+	end
+
+  def drop_outstanding_reviews(assignment_id)
+    emails =[]
+    reviews = ResponseMap.where(reviewed_object_id: assignment_id)
```
yangsong8: ResponseMap->ReviewResponseMap

app/models/drop_outstanding_reviews.rb
```diff
@@ -0,0 +1,21 @@
...
+    for review in reviews
+      review_has_began = Response.where(map_id: review.id)
+        if review_has_began.size.zero?
+          review_to_drop = ResponseMap.where(id: review.id)
```
yangsong8: line 14 is a waste, you can skip it and do line 15 as: `review.destroy`
This save one db query.
 But again the _review_ is a bad name..

app/models/drop_topic.rb
```diff
@@ -0,0 +1,34 @@
...
+         emails
+   end
+
+   def mail_sign_up_topic_users(assignment_id)
+       sign_up_topics = SignUpTopic.where(['assignment_id = ?', assignment_id])
+       assignment = Assignment.find(self.assignment_id)
+       emails =[]
+       for topic in sign_up_topics
+         signed_up_teams = SignedUpTeam.where(['topic_id = ?', topic.id])
+         unless assignment.team_assignment?
```
yangsong8: You should check the team size. If you check line 50 in assignment.rb, `assignment.team_assignment?` always returns true.

PR #837, opened 2016-11-05
E1633. Refactor different question types from quiz feature

psnakhwa:
Refactored different question types from quiz features. There were some issues in the first pull request, so we have created a new one. The original pull request had commits from all the collaborators.


yangsong8:
I tested:

- student_quizzes/review_questions page by clicking "view quizzes" icon. -->view_question_text works. But originally this method was defined in QuizQuestion. This should be a better place instead of putting 3 copied in 3 subclassess of QuizQuestion.
- questionnaires/view_quez by clicking "view" on "your work" page -->view_question_text works, the problem is the same.
- student_quizzes/finished_quiz. (by clicking "view" on a finished quiz, where grades should be shown) --> Does not display the finished quizzes correctly, view_completed_question does not work:
image

PR #824, opened 2016-10-30
E1635. Refactor join_team_requests_controller.rb and invitation_controller.rb

canhsiao:
## InvitationController
- Rename to **invitations_controller.rb,** in accordance with current naming convention.
- Convert if nested inside else to elsif.
- Add comments explaining what each method does, and comments on how important variables are used.
- Refactor create and accept methods. Shorten and clarify them by adding private methods (e.g. extract a ready_to_join? method and call it from accept method)
- Change awkward flash messages.
## JoinTeamRequestsController
- Add comments on some important variables.
- Convert if nested inside else to elsif
- Remove duplicate code: Common code in index, show and new method are moved into a private method. 
- Create and update do not share any code. This raises the spectre of DRY problems. Make sure that all functionality used in creating a request is available when editing a request, and that no code is duplicated.
- Fix wording of message:

PR #839, opened 2016-11-15
update - E1635. Refactor join_team_requests_controller.rb and invitation_controller.rb

canhsiao:
updated in accordance with last pull requests' comments.


app/controllers/invitations_controller.rb
```diff
@@ -8,63 +9,71 @@ def new
...
 
-  def update_join_team_request(user, student)
+  # update the request to join team 
+  # if join team request exist, handle invitaion automatically
+  # used in above case at line 51 when the condition is fulfiled
+  def update_join_team_request(user_invitee, student)
     # update the status in the join_team_request to A
-    if user && student
-      participant = AssignmentParticipant.where(['user_id =? and parent_id =?', user.id, student.parent_id]).first
+    if user_invitee && student
```
yangsong8: What is this `student` variable?

yangsong8:
It would be better if you tell me what changes have been made according to the comments from last pull request (especially, if you made changes for part of the issues but not all of them). As a matter of fact, I notice that no too much have been changed other than renaming.

app/controllers/invitations_controller.rb
canhsiao: This is the student in the existing team, as in line 14, we didn't modify this name because it looked reasonable by then. Now that you point out it's confusing, I think it does mean that the old naming needs improving.

canhsiao:
Hi Yang,

I apologize for the missing of information in the last commit. I replied on
each of the comments on the old commit.

https://github.com/expertiza/expertiza/pull/824/files

To explain some confusions here. For example, I believe there exists some
misunderstanding for the refactoring of invitation_controller.rb create
method, due to the indentation difference, the difference in the code
wasn't properly presented in github diff. For the file renaming, I'm not
sure why the change was lost in the final commit either. But this is indeed
our mistake. And I am sorry for the trouble it caused.

Please let me know if the comments are clear enough in explaining it. I
shall update further.

Thank you very much,
Zhongcan

P.S.

For the comment in new pull request, I updated a reply, too. I'm pasting it
below for your conveniece.

* yangsong8 <https://github.com/yangsong8> * a day ago
<https://github.com/expertiza/expertiza/pull/839/files#r88953008>
collaborator

What is this student variable?
This is the student in the existing team, as in line 14, we didn't modify
this name because it looked reasonable by then. Now that you point out it's
confusing, I think it does mean that the old naming needs improving.

canhsiao:
Hi, Yang,

I didn't see you reply, I'm wondering if the message didn't get to you
successfully.

Please let me know if the last message is clarifying enough.
Thank you very much!

Zhongcan

app/controllers/invitations_controller.rb
```diff
@@ -8,63 +9,71 @@ def new
...
+      participant = AssignmentParticipant.where(['user_id =? and parent_id =?', user_invitee.id, student.parent_id]).first
       if participant
         old_entry = JoinTeamRequest.where(['participant_id =? and team_id =?', participant.id, params[:team_id]]).first
         old_entry.update_attribute("status", 'A') if old_entry
       end
     end
   end
 
+  #auto complete user name
```
Winbobob: This comment is almost the same as the method name.

PR #846, opened 2016-11-26
Reset password email

shubham2892:


app/controllers/password_retrieval_controller.rb
```diff
@@ -12,13 +12,77 @@ def send_password
...
+     end
+  end
+
+  def reset_password
+    @email = params[:email]
+  end
+
+  def update_password
+    #method to be in user model
+    if params[:reset][:password] == params[:reset][:repassword]
```
shubham2892: Should we move this check to frontend and just use "password" here?

PR #853, opened 2016-12-02
E1689: Anonymous Chat Between Author and Reviewer

saratkavuru:
This is the version where reviewee would be able to chat with each reviewer in a separate pop-up box.


saratkavuru:
Please see #880 for the new request.

PR #854, opened 2016-12-02
E1690. Improvements to password recovery and repeated login failures

shubham2892:
Changes made:
- Google captcha after three failed attempt to login
- Send password reset link for resetting password instead of new password in mail.


app/controllers/auth_controller.rb
```diff
@@ -1,6 +1,8 @@
 class AuthController < ApplicationController
   helper :auth
 
+  @@attempts = 0
```
yangsong8: I got another idea which might be better than using a class variable. in your PasswordReset model, you can have a column "create_at". When a user request to reset the password, the code can check the # of reset requests done by the same user in, say, 24 hours. If the number is >=3, then doing recaptcha test is required.

shubham2892: @yangsong8 : What we were trying to achieve here was if user enters the wrong password for three consecutive times, then he will be shown with google recaptcha. This was to prevent from brute force attack.

yangsong8:
Below are comments for the grading purpose only. I will appreciate your clarification but please do not change code any more.

- I got this error message when I try to make login attemp the 3rd time:
image

Is there anything I missed other than "bundle install"?

spec/controllers/password_retrieval_controller_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  describe "check if token is expired" do
+    it "checks if token is not expired" do
+      local_token = "some random string"
+      @password_retrival = PasswordReset.new
+      @password_retrival.token = local_token
+      @password_retrival.user_email = "example@example.edu"
+      @password_retrival.save!
+
+      Timecop.freeze(Date.today + 2.days) do
```
yangsong8: If you mean the issue related to recaptcha, no, I still cannot have it run on my local machine. Is there anything I missed beside install the recaptcha gem?

shubham2892: @rsndates can you help him?

app/controllers/auth_controller.rb
anudeepvrm: but then won't that allow the script to brute force just by trying out different user names. He will get unlimited number of tries to make a script guess the password

spec/controllers/password_retrieval_controller_spec.rb
rsndates: @yangsong8 The issue with the counter may be that we used a class variable @@attemps instead of using an instance variable and adding a getter to the class.

rsndates: @yangsong8  if the counter is off by one attempt I would increment the count to be 4 or change the value of the initialized variable

shubham2892: @yangsong8 : I am not able to reproduce the issue locally, can you try including `include Recaptcha::Verify` in _application_controller.rb_. This [answer](http://stackoverflow.com/a/37960815/1754982) on stackoverflow discusses the same issue. I can also modify the code to include this, whatever is convenient for you.

PR #855, opened 2016-12-02
E1691 Convert Assignment Creation Form to ReactJS

KinshukBasu:


yangsong8:
Below are comments for grading purpose. Please do not change your code unless you are required to:

- A new pop-up panel when instructors click "new public assignment" or "new private assignment". But the logic on the popped up panel is not exactly the same as what user will see in when clicking edit icon on tree display, e.g. the user can use directory name "privated_assgt" on view/assignments/edit page, but cannot on the popup view. This means that we may have two separate logic on handling the same thing.
- When an instructor creates a new topic, Expertiza will bring the instructor to a new page.  But when the instructor edit the topic, it will be handled by ReactJS. This may add confusion.

PR #858, opened 2016-12-02
E1679. Let expert as well as students do reviews

dudu9929:
E1679


app/views/assignments/edit/_expert_review.html.erb
```diff
@@ -30,12 +30,60 @@
...
+              <%end%>
+          <% elsif @assignment.get_current_stage(topic_id) != "Complete" && @assignment.can_review(topic_id) %>
+
+                  <%= link_to "Begin", {:controller => 'review_mapping', :action => 'add_calibration', :id => assignment.id, :team_id => team.id} %>
+
+          <% else %>
+              Work has not yet been submitted for this submission
+          <% end %>
+
+            <% if false %>
```
yangsong8: What's this?

app/views/grades/_review_table.html.erb
```diff
@@ -4,7 +4,12 @@
   <%end%>
 	<tr><th style="text-align:center;">&nbsp;</th>
 		<% for i in 0...reviews.size%>
-			<th style="text-align:center;"><%= collabel %>&nbsp;<%=i+1%>&nbsp;</th>
+      <% user = User.find_by_id(AssignmentParticipant.find(ResponseMap.find(reviews[i].map_id).reviewer_id).user_id) %>
+        <% if(user.role.instructor? || user.role.ta?) %>
```
yangsong8: The logic here is not correct. One TA can still be a student in another assignment.

Winbobob: Actually, it is a design problem in Expertiza. Currently, `role_id` is in `User` table. In this case, if one student is TA of one assignment, s/he will be displayed as TA in all assignments.
 It will be better to put this field in `Participant` table.

PR #864, opened 2016-12-03
E1697. Add descriptions, links to topics

joshio1:


app/views/response/response.html.erb
```diff
@@ -29,7 +29,12 @@ jQuery(document).ready(function() {
         <% topic_id = SignedUpTeam.where(participant.parent_id, participant.user_id) %>
     <% end %>
     <% if !topic_id.nil? %>
-        <h2>You are reviewing <%= SignUpTopic.find(topic_id).topic_name %></h2>
+        <% topic = SignUpTopic.find(topic_id)%>
+        <h2>You are reviewing <%= topic.topic_name %></h2>
+        <b>Description &nbsp;&nbsp;&nbsp;</b>
+        <%= topic.description.to_s[0..30]+'...'%>
+        <a style="cursor: pointer" onclick="show_alert('<%= topic.description%>')">more...</a>
+        <%= link_to image_tag('external-link-16.png', :border => 0, :align => 'middle'), topic.link, :target=>'_blank' %>
```
yangsong8: You don't need to add another column, plz add this icon to topic name column

app/views/sign_up_sheet/_table_header.html.erb
```diff
@@ -10,6 +10,7 @@
   <th width="10%">Bookmarks</th>
   <th width="10%">Actions</th>
   <th width="5%">Advertise- ment(s)</th>
+  <th width="5%">Link</th>
```
yangsong8: could you make the link icon show next to "more..." as part of  "Topic name(s)" column. In this way a column will be saved. Thanks.

joshio1: Okay. I will try to do that.
I have handled the "30" size issue which you had earlier pointed out.
It is giving me some problem in CodeClimate which I am trying to figure out.

yangsong8:
Do you think the current version is ready to be graded? let me know when it is ready.

joshio1:
I have committed the latest changes which you have proposed. i.e. I have added links in the topic names instead of a link column and I have handled the error scenarios when description is >30.
The build request has also passed along with Code Climate.
I think it is ready to be graded.
Sorry for the delay. Thanks for your consideration.

yangsong8:
The comments below are for grading purpose only. Please do not change your code:

- description and url cannot be edited from UI. If an instructor want to make change on the description, (s)he has to re-import.

app/views/sign_up_sheet/_topic_names.html.erb
```diff
@@ -1,13 +1,25 @@
...
 <% else %>
-    <span class="inline" id="<%= i %>_show"><%= topic.topic_name %></span>
+    <span class="inline" id="<%= i %>_show"><%= topic.topic_name %> </span>
+<% end %>
+<% if !topic.description.nil? %>
+    <br/>
+    <% if topic.description.to_s.length>90 %>
+        <%= topic.description.to_s[0..90]+'...'%>
+        <a style="cursor: pointer" onclick="show_alert('<%= topic.description%>')">more</a>
+        </td>
```
yangsong8: This </td> cause the trouble. It closes the cell before the hyperlink icon.

app/helpers/import_topics_helper.rb
```diff
@@ -7,6 +7,13 @@ def self.define_attributes(columns)
     attributes["topic_name"] = columns[1].strip
     attributes["max_choosers"] = columns[2].strip
     attributes["category"] = columns[3].strip if columns.length > 3
+    define_attributes_extra(attributes, columns)
+    attributes
+  end
+
+  def self.define_attributes_extra(attributes, columns)
```
yangsong8: I don't like the code change here..

joshio1: Code Climate was giving me an issue about more number of Assignment Branch Condition size here. You can see that I had not done this earlier. I only did this after Code Climate complained about it.

PR #867, opened 2016-12-03
E1687. Instructor account creation over the web.

srikarpotta:


app/controllers/instructions_controller.rb
```diff
@@ -0,0 +1,9 @@
+class InstructionsController < ApplicationController
```
yangsong8: Why this controller is necessary?

app/controllers/users_controller.rb
```diff
@@ -71,6 +78,15 @@ def list
     @letters = ('A'..'Z').to_a
   end
 
+  def review
```
yangsong8: I dont think it's a good idea to have a review method in users_controller.

yangsong8:
The comment below is for grading purpose only. Do not change your code unless your are required to

app/views/users/request_new.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>Request new user</h1>
+
+<%= form_tag :action => 'request_user_create' do %>
+  <p><label for="role_id">Role</label><br/>
+    <%= collection_select 'user', 'role_id', @all_roles, :id, :name, {:selected => @rolename.id} %></p>
+
+  <%= render :partial => 'user', :locals => { :new_user => true, :request_user => true } %>
+  <%= recaptcha_tags %>
```
yangsong8: I got an error caused by this line when I deploy your code on my local machine.

app/views/mailer/request_user_message.html.erb
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+</head>
+<body>
+Hi <%= @first_name %>,<br />
+<br />
+Below user requested to sign up. Review by going to 'Manage/Users/Review Requested Users'.<br />
```
yangsong8: Cannot find this from UI.

app/controllers/users_controller.rb
chaitanya91k: Any suggestions on where to move this method. Should it be under requested_users_controller.rb?

app/controllers/instructions_controller.rb
chaitanya91k: hmm yes. Thanks for pointing out. As there are no additional functionality in this, we can remove this.

chaitanya91k:
For this -- **In the video, super admins can user menu item manage->user->Review Requested Users. But I do not see this in my local system.**

The cache in roles for super_Administrator in roles table needs to be refreshed. I could not figure out a clean way to refresh the cache. 
Well I deleted the cache from database roles table for super_Administrator and then reloaded the cache by logging in as a super_administrator (I had to comment few lines of code in models/menu.rb to load the cache)(Although I am sure there is a neat way to refresh the cache, I am not aware of).

**In the user/request_new page, there is no place for reason field in "Requested_user". In this the reason that the user wants to user Expertiza, or the reason that we reject the request? (I would miss something, but I didn't see this field is used either way.)**

Reason field was intended to be used bu Super_Administrator to log why they have rejected users request.
This is completely changeable based on what is requirement.   

**In the user/request_new page, a user has to select an institution from drop down.**

Yes, We are using the same view of new user. Shouldn't it be a drop down?
 We can change this if you need. If so, can you give more details on what exactly is required.

**No check on if the requested user name is available or no. We cannot allow an outside user to request an account "efg".**

We were planning to implement and thus added a TODO, But completely forgot about this. We can add this now if its ok.

PR #873, opened 2016-12-03
E1692 : Integrate Simicheck into Expertiza

kaps1489:


yangsong8:
Ferry says it does more damage than good. We should throw it away.

PR #877, opened 2016-12-03
E1708 Improvements to staggered-deadline assignments

SmokingSadhus:


db/migrate/20161202124659_add_review_last_graded_date_to_participants.rb
```diff
@@ -0,0 +1,5 @@
+class AddReviewLastGradedDateToParticipants < ActiveRecord::Migration
+  def change
+    add_column :participants, :review_last_graded_date, :datetime
```
yangsong8: should b called review_graded_at

Winbobob:
- I don't think add `review_graded_at` to `participants` table is correct. For newly-added reviews in staggered-deadline assignment, we can check the timestamp of responses.
 - I don't think add a new column in assignment editting page, due date tab is the right way to meet the 2nd requirement. I think what professor means is for topic deadlines instead of assignment deadlines.

Winbobob:
See dc9a90be9248900d8f70b013f6524ca669df9ed4

Winbobob:
See 93897ffe6a05d5369557c174300f6f1a6beb71c0

PR #880, opened 2016-12-03
E1689: Anonymous Chat Between Author and Reviewer

saratkavuru:
This is the version where the reviewee would be able to chat with all of their reviewers at once in a pop up box.


yangsong8:
You submitted two pull requests, #880 and #853, which is the right one, plz?

saratkavuru:
Hi Yang,
#853 is the version where each of  the reviewers have a separate chat with
the reviewee team.
#880 is the version where all of the reviewers have a common chat with the
reviewee team.
We have earlier done the #853 version and later after talking to the
professor , we have created #880 from another branch.
Please go through #880 first.

ᐧ

app/helpers/messages_helper.rb
```diff
@@ -0,0 +1,12 @@
+module MessagesHelper
+
+def reviewee_or_reviewer(message)
```
yangsong8: I don't like this method much. Better to have an 'is_reviewer' method?

app/helpers/messages_helper.rb
```diff
@@ -0,0 +1,12 @@
...
+	team = AssignmentTeam.find(message.chat.assignment_team_id)
+  for user in team.users
+    if message.user==user
+      return "Reviewee"
+    end
+      end
+	return "Reviewer"
```
yangsong8: White spaces don't look right to me.

app/helpers/messages_helper.rb
saratkavuru: @yangsong8  I have fixed this and changed the code to display Author instead of Reviewer in the chat box.

yangsong8:
Hi guys, I got trouble in running your code locally. Below is the error message I got. 

image

Do you have any guess on what I miss? Thanks.

saratkavuru:
Hi Yang, 
please run the following command **rails generate sync:install**
And then start the faye server by **rackup sync.ru -E production** along with **rails server**  in another shell.

yangsong8:
Thank you for your response!

saratkavuru:
Np, Did it solve the issue?

yangsong8:
Yes, I am moving on to test your project further. Yours is a big project considering the number of lines. I will check it carefully before making a decision. Thanks again.

yangsong8:
I got another issue. Below is what I have done.

- add a new participant.
- have the participant select a topic (thereby an assignment team is created and a chat object too)
- have a reviewer review the newly submitted work (chat object is created)

Now I am impersonating the review and try to say something. When I hit enter, I got an error on Rails console:
image

saratkavuru:
Hi Yang, we missed this usage instance of and old method. I have changed the method to its new name. Sorry for the trouble.

Winbobob:
I tried. The effect is different from your video. Each time after send a new message, I can see it only after reopen the chatting window.

PR #882, opened 2016-12-04
E1688: Issues on tree display fixed

VivekBhat:
Tree display for assignments had 1-2 lines of text, but always 3 lines of icons. This is a waste of space. 
Fixed this by adding a new single Icon that has the 3 lines of icons and pops up when clicked


db/migrate/20161126211944_create_feedback_supports.rb
```diff
@@ -0,0 +1,8 @@
+class CreateFeedbackSupports < ActiveRecord::Migration
```
yangsong8: The feedback to support team should not be saved into db.

app/views/shared/_footer_message.html.erb
```diff
@@ -1,2 +1,2 @@
-<a href="http://wiki.expertiza.ncsu.edu/index.php/Expertiza_documentation" target='_blank'>Help</a> | <a href="http://research.csc.ncsu.edu/efg/expertiza/papers" target='_blank'>Papers on Expertiza</a>
+<a href="http://wiki.expertiza.ncsu.edu/index.php/Expertiza_documentation" target='_blank'>Help</a> | <a href="http://research.csc.ncsu.edu/efg/expertiza/papers" target='_blank'>Papers on Expertiza</a> | <a href="/feedback_supports/new"> Expertiza yes Support</a>
```
yangsong8: Typo there.

yangsong8:
The comments you have received are for grading purpose only. Plz do not change you code unless you are told to.

PR #889, opened 2016-12-07
E1694

sandeeppilania:
Final Push : Completed Project

PR #890, opened 2016-12-07
E1694. Improve grade-conflict reports for instructors

sandeeppilania:


yangsong8:
The comments below are for grading purpose only. Please do not change your code unless you are required to.

- Two pull request #890 and #889.
- Pull request not named correctly.

app/controllers/response_controller.rb
```diff
@@ -181,6 +181,13 @@ def create
     msg = "Your response was successfully saved."
     error_msg = ""
     @response.email
+    begin
+      Response.compare_scores(@response, @questionnaire)
+
+      msg = "Your response was successfully saved."
```
yangsong8: At this point, the response record in db has already been created. So there's no point to rescue and give the error message saying "An error occurred..."’

app/controllers/response_controller.rb
```diff
@@ -181,6 +181,13 @@ def create
     msg = "Your response was successfully saved."
     error_msg = ""
     @response.email
+    begin
+      Response.compare_scores(@response, @questionnaire)
```
yangsong8: Does this mean that `compare_score` will be called even for a unsubmitted response?

app/models/response.rb
```diff
@@ -217,4 +217,83 @@ def self.get_volume_of_review_comments(assignment_id, reviewer_id)
...
+             :summary_url => 'https://expertiza.ncsu.edu/grades/view_team?id='+team_id.to_s #mapping.reviewee_id.to_s
+         }
+        }
+    ).deliver_now
+  end
+  def self.compare_scores(new_response, questionnaire)
+    map_class = new_response.map.class
+    existing_responses = map_class.get_assessments_for(new_response.map.reviewee)
+    total, count = get_total_scores_for_prev_reviews(existing_responses, new_response)
+    if count > 0
```
yangsong8: I don't think this is a good idea, this means what if this reviewee received more than 1 review from other reviewer, this will be triggered and 'notify_instructor' will be called.

app/models/response.rb
```diff
@@ -217,4 +217,83 @@ def self.get_volume_of_review_comments(assignment_id, reviewer_id)
...
+        |item|
+      if item.id != curr_item.id
+        count += 1
+        total += item.get_total_score
+      end
+    }
+    return total, count
+  end
+
+  def self.notify_instructor(assignment, curr_item, questionnaire, total, count)
```
yangsong8: This method does more (test the notification_limit) than the method name indicates.

app/mailers/mailer.rb
```diff
@@ -13,6 +13,11 @@ def generic_message(defn)
     @new_pct = defn[:body][:new_pct]
     @avg_pct = defn[:body][:avg_pct]
     @assignment = defn[:body][:assignment]
+    @limit = defn[:body][:limit]
```
yangsong8: arguably, the messages that this project send is not generic message. It would be better to create another method, e.g. review_conflict_message.

app/models/response.rb
yangsong8: There could be a tricky issue caused by this: in Expertiza, reviewers do not need to answer all the rubric questions. If they skip some, the skipped questions will be omitted when calculating review scores. But if you do `get_maximum_score`, those question skipped by reviewer will be considered as answered with 0 as score. [Welp, this is not your bad. You need to know a lot about this project to be aware of this...]

app/models/response.rb
```diff
@@ -217,4 +217,83 @@ def self.get_volume_of_review_comments(assignment_id, reviewer_id)
...
+    max_possible_score = curr_item.get_maximum_score
+    new_score = curr_item.get_total_score;
+    existing_score = (total.to_f/count).to_f
+
+    aq = AssignmentQuestionnaire.where(user_id: assignment.instructor_id, assignment_id: assignment.id, questionnaire_id: questionnaire.id).first
+
+    unless aq
+      aq = AssignmentQuestionnaire.where(user_id: assignment.instructor_id, assignment_id: nil, questionnaire_id: nil).first
```
yangsong8: Not sure if the assignment_id can be nil here.

app/models/response.rb
```diff
@@ -217,4 +217,83 @@ def self.get_volume_of_review_comments(assignment_id, reviewer_id)
...
     [overall_avg_vol, avg_vol_in_round_1, avg_vol_in_round_2, avg_vol_in_round_3]
   end
+  def notify_on_difference(new_pct, avg_pct, limit,response_id)
+    mapping = self.map
+    instructor = mapping.assignment.instructor
+    team_id = mapping.reviewee_id
+    teams_users = TeamsUser.where(team_id: team_id)
+    user_id=nil
+    teams_users.each do |team_user|
+      user_id=team_user.user_id;
```
yangsong8: I don't understand the code here. You loop all the users in reviewee team, then what?

PR #891, opened 2016-12-07
E1696 IMPROVE SELF REVIEW

cvvishal:
vchitto
mmurali5
nnatara2


app/controllers/grades_controller.rb
```diff
@@ -88,6 +88,27 @@ def view_my_scores
     @summary = sum.summary
     @avg_scores_by_round = sum.avg_scores_by_round
     @avg_scores_by_criterion = sum.avg_scores_by_criterion
+
+
+
+
+    @review_mappings = SelfReviewResponseMap.where(reviewer_id: params[:id])
```
yangsong8: The variable naming is not correct. @review_mappings variable is not a response map, but a SelfReviewResponseMap.

app/controllers/grades_controller.rb
```diff
@@ -88,6 +88,27 @@ def view_my_scores
...
+    @sorted_responses = Array.new
+    @prev = Response.where(:map_id => @review_mappings[0].id)
```
yangsong8: Will there b an error if the query on line 95 get nothing?

cvvishal: If line 95 returns nothing, then no self review exists. If no self review exists then this page wont be invoked

app/controllers/grades_controller.rb
```diff
@@ -88,6 +88,27 @@ def view_my_scores
...
+    array_not_empty = 1
+    @sorted_responses << element
+    end
+    if (array_not_empty == 1)
+      @sorted_responses = @sorted_responses.sort_by { |obj| obj.updated_at }
+      @latest_response = @sorted_responses.last
+    end
+
+    @response = Response.find(@latest_response.id)
+    $max_score=@response.get_maximum_score
```
cvvishal: Have changed the global variable to local variable

PR #892, opened 2016-12-07
E1677: Text Metrics Project

sequae92:
This is part of adding Text Metrics as a tool for the instructor and the student to check their reviews.


yangsong8:
Below are some comments for grading purpose only. 

- video submitted to file server.

app/models/review_metric.rb
```diff
@@ -0,0 +1,134 @@
...
+                    is_suggestion = true
+                end
+                if problem_words.include? word
+                    is_problem = true
+                end
+            end
+
+            #diff_word_count = response_level_comments[current_response_id].scan(/[\w']+/).uniq.count
+            if ReviewMetric.exists?(response_id: key)
+                obj = ReviewMetric.find_by(response_id: key)
```
yangsong8: So review metrics is only calculated when the instructor visit this information ,right?

sequae92:
Dear Song,

Yes, we followed a approach where the instructor has been kept in control
of the review metrics in this case. Students will be able to see it after
the instructor has viewed the report once.

Thanks and Regards,
Spondon

PR #893, opened 2016-12-11
E1703: Additions to Logging in Expertiza

AswinAK:
Pull request for our final project.


app/helpers/questionnaire_helper.rb
```diff
@@ -1,99 +1,24 @@
 # OSS808 Change 28/10/2013
 # FasterCSV replaced now by CSV which is present by default in Ruby
 # require 'fastercsv'
-# require 'csv'
+ require 'csv'
```
yangsong8: What's the reason that you change this file?

app/models/questionnaire.rb
```diff
@@ -92,4 +93,37 @@ def validate_questionnaire
     results = Questionnaire.where(["id <> ? and name = ? and instructor_id = ?", id, name, instructor_id])
     errors.add(:name, "Questionnaire names must be unique.") if !results.nil? and !results.empty?
   end
+
+  def to_csv(ques)
```
yangsong8: This is not related to the final project, right?

AswinAK: No, this is not related to our project

app/helpers/questionnaire_helper.rb
AswinAK: This was done by mistake, not part of our changes.

PR #544, opened 2015-04-24
E1529. Extend the Email notification feature to scheduled tasks

KeleiAzz:
Final project E1529

PR #897, opened 2017-01-23
E1529 Extend the Email notification feature to scheduled tasks

yangsong8:


yangsong8:
This is a duplicated. See #544

PR #900, opened 2017-01-30
E1690. Improvements to password recovery and repeated login failures

ferryxo:
Bugs fixed by Ferry and ready to merge


ferryxo:
closing pull request to trigger travis CI build

yangsong8:
Below are some of my comments for this project (no for you, my buddy). Nothing major, though

- If login fails once, the system will redirect the user to expertiza.ncsu.edu/auth/login and provide another login page. This is the page supports recaptcha. This page is potentially a wasted because they can just add recaptcha to expertiza.ncsu.edu page.
image
- The number of attempts does not increase if I skip recaptcha and just type in a wrong password again, which is a pretty obvious mistake.
- After requesting to reset a password, the program should render user to expertiza.ncsu.edu page rather than staying on password reset page.
image

ferryxo:
>>The number of attempts does not increase if I skip recaptcha and just type in a wrong password again, which is a pretty obvious mistake.

no this is not a mistake, if the user doesn't do the captcha, the system shouldn't even try to login. because captcha is used to avoid brute force. if the system tried to login, it allows users to bypass captcha and do brute force. 
maybe the error msg needs to be revised

PR #912, opened 2017-03-22
E1726.  Remove cache field in roles table

gokkulasudanr92:
Certain model files, e.g., factories.rb, contain long text strings with the text used in menus (e.g., “Users \n Questionnaires \n Courses ... “ above).  Storing such large strings in the database may not be a good idea. It is because if you want to change the large strings, you need to write a migration, which is quite inconvenient. Actually, you can hack it by deleting the certain lines from “cache” field directly. However, the potential problem is that it will make databases used by different users inconsistency. If we can store these large strings in YAML files (each role one file), it will be easier to make modifications.


Winbobob:
Since you change the DB structure, please also commit `db/schema.rb
`

gokkulasudanr92:
We have changed the db/schema.rb and also 20170317212555_remove_cache_field_from_roles.rb in migration has also been added.

spec/models/role_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+require 'yaml'
+require 'json'
+
+expected_student_cache = YAML.load_file("#{Rails.root}/config/role_student.yml")[Rails.env]
+expected_instructor_cache = YAML.load_file("#{Rails.root}/config/role_instructor.yml")[Rails.env]
+expected_admin_cache = YAML.load_file("#{Rails.root}/config/role_admin.yml")[Rails.env]
+
+describe Role do
+  before :all do
+    @student_role = build(:role_of_student, id: 1, name: "Student", description: '', parent_id: nil, default_page_id: nil)
```
Winbobob: Use `build` instead of `create` here is very good. 👍

app/models/role.rb
```diff
@@ -2,15 +2,28 @@
 require "menu"
 
 class Role < ActiveRecord::Base
+  include ApplicationHelper
```
Winbobob: Include `ApplicationHelper` in model file is not a good idea, there should be some other ways to do it.

Winbobob:
Merge after refactoring.

db/schema.rb
```diff
@@ -433,7 +433,6 @@
     t.integer  "parent_id",       limit: 4
     t.string   "description",     limit: 255,   default: "", null: false
     t.integer  "default_page_id", limit: 4
-    t.text     "cache",           limit: 65535
```
Winbobob: I think I know the reason why the travis ci failed. Do you manually delete L436? If so, it is not correct. You have to commit the `schema.rb` after running migrations.

PR #922, opened 2017-03-27
E1712. Refactor join_team_requests_controller.rb and invitation_controller.rb

Winbobob:


Winbobob:
Hi team,

Please try to fix the issues reported by code climate.

Thanks,
Zhewei

kira0992:
Hi Zhewei

We are not able to figure out why codeclimate is throwing up the inconsistent indentation issue. Could you help us out. 

Thanks

ruthvik1994:
Hi,
3 of the issues with High ABC size are described as 20/15, 19/15 and 17/15. We couldn't reduce it further. 
And the other 2 issues involves just a similar redirect_method line (Duplicity)
We feel it is not required to put redirect into another method and call it
Need any advice if any to overcome this issue

Thanks

yangsong8:
The comments below are for grading purpose only, please do not change your code unless you are told to.

Edit join team request function is broken:
image
image

kira0992:
Hi Yang

We did not change that function. It must have been broken before we took it up. None of the tasks given to us had anything to do with editing an existing join team request, so we never got around to checking that functionality. We were only dealing with creating a new join team request as per the issues identified in the document.

We will fix it now if you want us to, just wanted to let you know.

Thanks 
Aravind

app/controllers/join_team_requests_controller.rb
```diff
@@ -90,9 +66,31 @@ def destroy
...
   end
+
+  private
+
+  def check_team
+    # check if the advertisement is from a team member and if so disallow requesting invitations
+    team_member = TeamsUser.where(['team_id =? and user_id =?', params[:team_id], session[:user][:id]])
+    team = Team.find(params[:team_id])
+
+    return flash[:note] = "This team is full." if team.full?
```
yangsong8: The logic here is not correct. One can still request, and the request can be saved with a flash message.

See the example below:


class ApplicationController < ActionController::Base
  before_action :require_login
 
  private
 
  def require_login
    unless logged_in?
      flash[:error] = "You must be logged in to access this section"
      redirect_to new_login_url # halts request cycle
    end
  end
end

yangsong8:
No, we will handle those issues if we plan to merge this pull request. You guy can focus on final project now.

app/helpers/invitations_helper.rb
```diff
@@ -0,0 +1,2 @@
+module InvitationsHelper
```
yangsong8: Maybe we don't need this.

PR #923, opened 2017-03-27
E1716. Improve e-mail notifications

manoj4269:


Winbobob:
Hi team,

Please revert the changes on `rubocop.yml` and `.travisci.yml`.

Thanks,
Zhewei

app/controllers/invitation_controller.rb
```diff
@@ -9,8 +9,11 @@ def new
 
   def create
     user = User.find_by_name(params[:user][:name].strip)
+
     team = AssignmentTeam.find(params[:team_id])
     student = AssignmentParticipant.find(params[:student_id])
+    student_email = student.user_id
```
efg: What?  User_ID normally doesn't contain the domain, but email needs to.  Sounds like a bad variable name, or worse.

app/controllers/invitation_controller.rb
```diff
@@ -37,6 +40,7 @@ def create
             @invitation.assignment_id = student.parent_id
             @invitation.reply_status = 'W'
             @invitation.save
+            accept_invitation(student_email)  # Call the mailing function
```
efg: The mailing function is called accept_invitation?  This is really confusing.

app/controllers/invitation_controller.rb
```diff
@@ -65,8 +79,20 @@ def auto_complete_for_user_name
     @users = User.where("LOWER(name) LIKE ?", "%#{search}%") unless search.blank?
   end
 
+  def accepted_invitation(user_email)
```
efg: Another badly named method.  Should be email_an_acceptance

manoj4269: Hello Prof, we have fixed all the naming conventions that you have suggested us.
accept_invitation => email_an_invitation
accepted_invitation => email_an_acceptance
student_email => student_id

manoj4269:
Hello Zhewei
We have refactored and cleared out the issues in the code climate. But some of the issues showing up are from the code written earlier in some other controllers which we haven't touched at all and some are from the mailer which was also present earlier and can't be changed. 
Thanks, 
Manoj

Gemfile
```diff
@@ -66,7 +67,6 @@ group :development do
   gem 'pry'
   gem 'pry-remote'
   gem 'pry-nav'
-  gem 'rubocop'
```
Winbobob: Why do you delete this gem?

app/helpers/import_file_helper.rb
```diff
@@ -18,6 +18,9 @@ def self.create_new_user(attributes, session)
     user.parent_id = (session[:user]).id
     user.timezonepref = User.find(user.parent_id).timezonepref
     user.save!
+    password = 'qwerty' # the password is reset
```
Winbobob: I remember the password is auto-generated instead of hard coded after importing new students.

PR #931, opened 2017-04-02
E1720. UI issues/fixes

rmmaily:
This is the work done by our group


rmmaily:
So people can check.

db/schema.rb
```diff
@@ -287,6 +288,15 @@
     t.string  "type",           limit: 255
   end
 
+  creatications", force: :cascade do |t|
```
Winbobob: This line seems not correct for me.

Gemfile
```diff
@@ -8,9 +8,9 @@ gem 'activerecord-session_store'
 gem 'airbrake'
 gem 'awesome_nested_set'
 gem 'bind-it'
-gem 'capistrano'
```
Winbobob: You do not need to change `Gemfile` and `Gemfile.lock`.

PR #937, opened 2017-04-11
E1733 Text Metrics

rmmaily:
Initial setup with new


Winbobob:

app/views/review_mapping/response_report.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= render :partial => 'searchbox' %>
```
Winbobob: Why did rewrite `haml` view back to `erb` view?

config/text_metrics_keywords.yml
```diff
@@ -0,0 +1,14 @@
+development:
```
Winbobob: IMHO, these keywords could be the same in different environments. Hence, I think you do not need to duplicate them twice.

spec/features/heat_map_spec.rb
```diff
@@ -60,28 +60,6 @@ def create_review
     expect(page).to have_content('Summary Report for assignment')
   end
 
-  it 'should be able to follow the link to a specific review' do
```
Winbobob: Why did you delete these two test cases? Or it is because they are flaky tests (sometimes pass, sometimes fail)?

Winbobob:
- It will be better to have a demo video.
- Please fix `code climate` issues as many as possible.

db/seeds.rb
```diff
@@ -0,0 +1,4 @@
+ReviewMetric.create!(metric: 'Volume')
```
Winbobob: You may add the content to [factories.rb](https://github.com/expertiza/expertiza/blob/master/spec/factories/factories.rb).

app/controllers/popup_controller.rb
```diff
@@ -119,4 +146,41 @@ def reviewer_details_popup
...
+  def calculate_percentages(assign_id)
+    #single_response = Response.where(id: record_id)
+    #mapped_response = ResponseMap.where(id: single_response[0].map_id)
+    review_maps = ResponseMap.where(reviewed_object_id: assign_id)
+    keys = [[0.00, 0.00, 0.00, 0.00], [0.00, 0.00, 0.00, 0.00], [0.00, 0.00, 0.00, 0.00], [0.00, 0.00, 0.00, 0.00]]
+    response_count = [0.00, 0.00, 0.00, 0.00, 0.00, 0.00]
+    word_counter = [0, 0, 0, 0, 0, 0]
+    suggestive_count = [0, 0, 0, 0, 0, 0]
+    problem_count = [0, 0, 0, 0, 0, 0]
+    offensive_count = [0, 0, 0, 0, 0, 0]
```
yangsong8: It seems to me that if we want to add another other metrics, say, unique words, we have to change the code here. Is there any metaprogramming can be possibly used here?

PR #938, opened 2017-04-13
Review grader enhancements - Text Macros

joshio1:
Students, Reviews, and Instructors several times have to type the same feedback or comment on several occasions. Text Macros help in typing the same content over and over again. User can save a combination of short-form and long-forms and then while he/she is typing, the short-forms will be automatically replaced by the long-form on a special character key-press.


Winbobob:
Instead of add javascript files under `vendor`, we can add it to bower.

Winbobob:
It is a little bit weird that when first time loading report view, there is a form at the end of the view.

joshio1:
Hi Zhewei,

Thanks for pointing out.
I'll check on my environment and fix it.

Thanks and Regards,
Omkar Joshi

joshio1:
Hi Zhewei,
I have handled the case where the form was being shown in the first load.
Also, as requested,
awesome_input.js - javascript file which I wrote by myself.
awesomplete.js - javascript file from the awesomplete library which I modified for our requirements.

Winbobob:
LGTM

PR #915, opened 2017-03-23
E1729. Export scores in detail

kkushagra:
Changes to add new feature, which exports a detailed CSV.

The changes have been explained in wiki : http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Spring_2017/oss_E1729

PR #939, opened 2017-04-25
Revert "E1729. Export scores in detail"

Winbobob:
Reverts expertiza/expertiza#915
Reason: `Not following the existing design`. 
Below is the correct design.
qq


yangsong8:
This pull request is reverted because of floppy code, especially, ~150 lines of code got added to assignment.rb.

PR #943, opened 2017-04-28
E1744.  Github Metrics Integration

bigfatnoob:


Winbobob:

bigfatnoob:
@Winbobob: I think for the the CI to pass db:migrate has to be run I guess

db/migrate/20170421194857_create_github_contributors.rb
```diff
@@ -0,0 +1,18 @@
+class CreateGithubContributors < ActiveRecord::Migration
+  def change
+    create_table :github_contributors do |t|
```
Winbobob: It will be better to create a `github_metrics` table first and then use `metics_id` as the foreign key in `github_contributors` table. In this way, if we want to add more metrics in the future, we only need to add a new record in `github_metrics` table, instead of changing db structure.

Winbobob:
@bigfatnoob, since you added a new table, you have to commit the `schema.rb` file. Travis CI will run `db:migrate`.

db/migrate/20170421194857_create_github_contributors.rb
bigfatnoob: The `github_contributors` table has privilege to add additional columns to it. We can do the change you suggested but this will make the insert very slow since now we will be inserting in two tables. Do you think we should still make these changes?

Winbobob: You can keep the current design. I just think about some other ways to do it.

spec/controllers/github_contributors_controller_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'rails_helper'
+include LogInHelper
+
+describe GithubContributorsController do
+  before(:each) do
+    student.save
+    instructor.save
+    @role = Role.new(name: 'Instructor')
```
Winbobob: Use `New` to create a new instance is good. We can also use `build` keyword.

spec/controllers/github_contributors_controller_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    @assignment_team = AssignmentTeam.where(name: 'My Github Team').first ||
+        AssignmentTeam.new(name: 'My Github Team', parent_id: @assignment.id)
+    @assignment_team.save
+    @sub_record = SubmissionRecord.new(content: 'https://github.com/mozilla/geckodriver',
+                                       operation: 'Submit Hyperlink',
+                                       team_id: @assignment_team.id,
+                                       user: @student.name,
+                                       assignment_id: @assignment.id)
+    @sub_record.save
+    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(@instructor)
```
Winbobob: Actually, these 3 lines have been implemented in `stub_current_user` in [rails_helper.rb](https://github.com/expertiza/expertiza/blob/master/spec/rails_helper.rb#L69)

yangsong8:
I don't think this is the right time to merge this project. The biggest issue for your project is that it pulls all the commits of the entire project, not just the commits done by this pull request. (Well, this means that Zhewei, Dr. Gehringer and I will be on the top of the list) To handle this, the potential extensions could be 1) put a limit of the time span or 2) have each team member submit their user name on Github.

But if you can use API which pulls only the commits of each pull request instead of the commits for the entire project, things could be much easier.

PR #944, opened 2017-04-28
E1731: Improve score calculation

psvaidya:
Changed OnTheFlyCalc module into a class and all its methods into static methods. Made the required changes wherever these methods are called.
Created new LocalDbFlyCalc class which stores total scores for each response map for an assignment in new "local_db_scores" table and sets new attribute "local_scores_calculated" in assignments table to true. 
When calling compute_total_scores method, if local_scores_calculated is set (true), then the method in LocalDbFlyCalc is called, else the method in OnTheFlyCalc is called.
New icon is added for each assignment on Manage Assignments page. Clicking on this icon stores the total scores in database by calling the method in LocalDbFlyCalc class.

For more details, check the wiki page: http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Spring_2017/E1731_Improve_Score_Calculation


Winbobob:

psvaidya:
We have fixed all the codeclimate issues except 2:
Assignment Branch Condition size: Tried to fix this by using private methods
Mass assignment: Fixed other similar issues using attr_accessible, but 1 still remained

Travis CI build passed in the earlier run but giving timeout error now, which is probably an issue with the server running that test. In details, it is showing build has passed.

yangsong8:
The comments below are for grading purpose only, please do not change your work unless you are told to:

- I calculate all the scores and store them to the local_db_scores table for assignment 765. For response_map id 111376, the reviewer did both round of review, but the first round of review score was calculated as 0.
image
image

psvaidya:
Hi Yang,
We found the error in the code because of which the round 1 scores are coming 0.

**models/local_db_cal.rb file line number 28**: _@response = Response.where(map_id: map.map_id).last_
Here, always the last response is taken which is the response of the latest round. So, the round number is also required in the where condition. Adding round number to the where condition fixes the issue

**Line number 28 NEW CHANGE**: _@response = Response.where(**round: round,** map_id: map.map_id).last_

Also, with this, 1 more change is required to check if the response is nil, otherwise it gives an error while checking if the response is submitted.
**Line number 31 NEW CHANGE**: _@response = nil unless **@response.nil?** or @response.is_submitted_

I have tested this on my local machine and with the above changes, round 1 scores are also calculated and stored correctly. I have not committed this change as you have asked not to make any changes. So just mentioned the solution here in the comment.

Regarding response_map_id 111411, I believe the score saved should be null, because the "is_submitted" column in that response table is false (0). This means the response is saved, but not submitted. So the expected score to be stored should be NULL.

psvaidya:
We have changed local_scores_calculated column name to local_scores_stored in Assignments table and committed it as requested by the professor.

test/fixtures/local_db_scores.yml
```diff
@@ -0,0 +1,13 @@
+# Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
```
Winbobob: This fixture is auto-generated. You may remove this file from pull request.

test/models/local_db_score_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
```
Winbobob: This fixture is auto-generated. You may remove this file from pull request.

app/models/participant.rb
```diff
@@ -124,7 +124,12 @@ def scores(questions)
       end
     end
 
-    scores[:total_score] = assignment.compute_total_score(scores)
+    # E1731 changes: Calling either LocalDbCalc or OnTheFlyCalc based on assignment status
+    scores[:total_score] = if self.assignment.local_scores_stored?
+                             LocalDbCalc.compute_total_score(assignment)
```
yangsong8: As discussed in Expertiza meeting May 9th, this should be done by metaprogramming.

PR #945, opened 2017-04-28
E1740. Lightweight badging system based on Credly

gokkulasudanr92:
The aim of this project is to implement a simple badging system for expertiza. You may use credly to create the badges. The badges can be earned automatically (for example: by having one of the top 3 scores in class)


Winbobob:

app/models/assignment.rb
```diff
@@ -622,4 +622,11 @@ def find_due_dates(type)
     self.due_dates.select {|due_date| due_date.deadline_type_id == DeadlineType.find_by_name(type).id }
   end
 
+  def self.first_submission_reward(ass_id, team_id)
```
Winbobob: Change variable name to `assgt_id`.

app/models/team.rb
```diff
@@ -75,6 +75,11 @@ def self.size(team_id)
     TeamsUser.where(["team_id = ?", team_id]).count
   end
 
+  # Check if team was first submitter on assignment
+  def self.first_sub(team_id, assignment_id)
```
Winbobob: Change this method name to `first_submission`.

app/models/team.rb
```diff
@@ -256,5 +261,12 @@ def self.create_team_and_node(id)
     team
   end
 
-  # REFACTOR END:: class methods import export moved from course_team & assignment_team to here
+  def self.get_team(participant)
```
Winbobob: I don't think you need to write a method to get the team info from a participant. 
You can try `Team.find_by(TeamsUser.team_id(participant.parent_id, participant.user_id) ||= 0)`
Also, calling a model method from view is not a good idea. It will be better to move the method to corresponding helper file if you want to keep this method/

Winbobob:
- Remove conflicts in `assignment.rb` and fix `code climate` issues as many as you can.
- For review average score, I understand the reason you store them in DB. However, for `first_sub_teamid`, I do not think it is necessary to store them in DB. It can be fetched by one query I think.

app/models/assignment.rb
```diff
@@ -622,4 +622,11 @@ def find_due_dates(type)
     self.due_dates.select {|due_date| due_date.deadline_type_id == DeadlineType.find_by_name(type).id }
   end
 
+  def self.first_submission_reward(assgt_id, team_id)
+    assignment = Assignment.find(assgt_id)
+    if !assignment.nil? && assignment.first_sub_teamid == -1
+      assignment.first_sub_teamid = team_id
```
yangsong8: I doubt this design: you added a "first_sub_team_id" in the assignment table. This means that if we r gonna extend this badge from "first submission" to "early birds submission" to give multiple teams this badge, we have to add more than 1 column to the assignments table.

I would suggest create a badges_mapping table which records assignment_id, team_id and badge name (or badge id) in order to make this feature more extendable.

Winbobob: We already have `submission_records` table. We can use that table to check 1st submission or early birds submissions.

app/models/review_response_map.rb
```diff
@@ -196,4 +196,9 @@ def self.prepare_review_response(assignment, maps, review_final_versions, round)
     end
     review_final_versions[symbol][:response_ids] = response_ids
   end
+
+  def self.good_reviewer?(participant)
+    score = participant.try(:review_grade).try(:grade_for_reviewer)
+    !score.nil? && score > 95
```
yangsong8: score && score > 95

Winbobob:
## Feedback
The badging functionality works. However, the design may not be extendable. You store the scores in different tables, which will accelerate the page loading time. However, a better way to achieve badging functionality is to create a `badges_mapping` table which records `assignment_id`, `participant_id` and `badge name (or badge id)` in order to make this feature more extendable. Also, add a `badges_controller.rb` to control the logic.

PR #946, opened 2017-04-28
E1742. Visualizations for instructors

ApertureScienceInnovators:
We've added new functionality for instructor visualization.  There's a new chart on the View Grades page that shows in a quick graphical form how the entire class did on an assignment (including across multiple rounds) and a new table that includes all relevant information in one place.


Winbobob:
- Hi team, please fix `code climate` issues as much as you can. Thx!
- We install `highcharts` javascript via `bower`, please remove those files from `vendor` folder.

ApertureScienceInnovators:
We've already fixed the CodeClimate issues as best as we can; we've gotten it down from ~55 to 2 but aren't sure how to deal with the remaining ones. 

For Highcharts -- do we simply delete the new files in the vendor folder or is there something else required?

Thanks!

Winbobob:
I think you can try to delete the files related to `highcharts` in `vendor` folder and make sure `bower install` has been executed successfully. In theory, no further actions are required. You can try the function on your local machine.

ApertureScienceInnovators:
Thank you.  We will check and make sure that it works and then push the changes if it does and get back to you if it does not.

ApertureScienceInnovators:
I've removed the spare files.  Thank you for the pointer.

Gemfile
```diff
@@ -27,6 +27,7 @@ gem 'jquery-rails'
 gem 'jquery-ui-sass-rails'
 gem 'jquery-datetimepicker-rails'
 gem 'json', '~> 1.8', '>= 1.8.3'
+gem 'lazy_high_charts'
```
Winbobob: - Actually, you do not need to add this gem but it is not your fault since we did not mention clearly in requirement.
- The `highchart` javascript library has already been added in Expertiza. See [link].(https://github.com/expertiza/expertiza/blob/master/bower.json#L31)

app/assets/javascripts/application.js
```diff
@@ -28,6 +28,12 @@
 //= require jquery.datetimepicker
 //= require jquery-tablesorter
 
+//= require_tree .
```
Winbobob: You may revert the changes in `application.js`.

app/controllers/grades_controller.rb
```diff
@@ -238,6 +253,21 @@ def get_body_text(submission)
...
 
+  def calculate_review_questions(assignment, questionnaires)
+    min = 0
+    max = 5
+
+    number_of_review_questions = 0
+    questionnaires.each do |questionnaire|
+      next unless assignment.varying_rubrics_by_round? && questionnaire.type == "ReviewQuestionnaire" # WHAT ABOUT NOT VARYING RUBRICS?
+      number_of_review_questions = questionnaire.questions.size
+      min = questionnaire.min_question_score < min ? questionnaire.min_question_score : min
```
yangsong8: This is not necessary, the information is stored in questionnaires table.

Winbobob:
## Feedback
 - For links in each cell. We are not sure what will happen if we click the link and which page will open. This part did not show in the video. 
 - For visualization part
   - The scale is blue to green, which does not make sense. And colors will change randomly each time loading the page. It will be better to scale from red to green. 
   - And two adjacent bar represents the response in round _1_ to round _k_. It makes sense only if the rubrics in all review rounds are all the same. If the instructor implements the vary-rubric-by-round mechanism, this visualization will not make sense. 
   - We may think about another visualization that hopefully can handle all different scenarios. 
 - The table is presorted by teams, but you can now also sort alphabetically. The cell view looks way too long, and should be divided into partials.

Winbobob:
I closed this pull request. It is because another final project (#947)  also worked on refactor `grades#view` page. We will make a decision in close future (use yours, use that one, or both).

PR #948, opened 2017-04-29
E1734: Improve Survey Functionality

sid5788:


app/models/questionnaire.rb
```diff
@@ -34,10 +34,12 @@ def get_weighted_score(assignment, scores)
                          'Teammate ReviewQuestionnaire',
                          'TeammateReviewQuestionnaire',
                          'SurveyQuestionnaire',
+                         'AssignmentSurveyQuestionnaire',
+                         'Assignment SurveyQuestionnaire',
```
yangsong8: I dont understand this

config/routes.rb
```diff
@@ -453,6 +455,7 @@
     collection do
       get :list
       get :delete
+      post :delete # change
```
yangsong8: ?? what does this "change mean"

db/migrate/20170428195326_replace_course_evaluation.rb
```diff
@@ -0,0 +1,18 @@
+class ReplaceCourseEvaluation < ActiveRecord::Migration
+  def up
+    Questionnaire.where(type: "CourseEvaluationQuestionnaire"){
+        | questionnaire |
```
Winbobob: If the block is multiline, it will be better to write the block like this
```
CourseEvaluationQuestionnaire.all do | questionnaire |
[code]
end
```

app/views/response/response.html.erb
```diff
@@ -18,41 +18,45 @@ jQuery(document).ready(function() {
 })
 </script>
 
-<h1><%= @header %> <%= @title %> for <%= @map.assignment.name %></h1>
+<% if @map.survey? %>
```
yangsong8: Is this necessary?

PR #949, opened 2017-04-30
E1738. Integrate Simicheck Web Service into Expertiza

djgutier:


Winbobob:

ferryxo:
Also please fix Travis CI until it's able to build. thx

djgutier:
Hi, we fixed all of the Code Climate issues that were reasonable.  The remaining ones will break implementation in order to resolve.

djgutier:
Just re-ran Travis CI and the build passed

app/controllers/plagiarism_checker_assignment_submissions_controller.rb
```diff
@@ -0,0 +1,2 @@
+class PlagiarismCheckerAssignmentSubmissionsController < ApplicationController
```
ferryxo: I'm not sure why you commit an empty class? remove this?

app/helpers/plagiarism_checker_helper.rb
```diff
@@ -0,0 +1,115 @@
+module PlagiarismCheckerHelper
+  require 'simicheck_webservice'
+  require 'submission_content_fetcher'
+
+  # PlagiarismCheckerHelper acts as the integration point between all services and models
```
ferryxo: shouldn't the integration be done in a controller?

app/views/assignments/edit/_general.html.erb
```diff
@@ -163,6 +163,20 @@ function microtaskChanged() {
 
     </td>
   </tr>
+  <tr>
+    <td style='padding:5px' id='assignment_simicheck_field'>
+      <%= label_tag('assignment_form[assignment][simicheck]', 'SimiCheck Delay') %>
+      <%= select('assignment_form[assignment]', 'simicheck', (-1..100).to_a.map{|n| n== -1? ['No SimiCheck', -1] : [n, n]}, :selected => @assignment_form.assignment.simicheck) %><b> hours</b>
```
yangsong8: "assignment.simicheck" is a bad name

spec/models/google_doc_fetcher_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    expect(GoogleDocFetcher.supports_url?("https://docs.google.com/document/d/1-waDJtPB8VGLyubb40-X951PF9R3jdhm6Kb4ga__S0E/edit")).to be_truthy
+  end
+
+  it "fetches a google doc export" do
+    http_setup_get_request_mock_success
+
+    params = {"url" => "https://docs.google.com/document/d/1-waDJtPB8VGLyubb40-X951PF9R3jdhm6Kb4ga__S0E/edit"}
+    fetcher = GoogleDocFetcher.new(params)
+    res = fetcher.fetch_content
+    expect(!res.empty?)
```
Winbobob: There should be some expectations, right?

app/controllers/plagiarism_checker_assignment_submissions_controller.rb
djgutier: Deleted

app/helpers/plagiarism_checker_helper.rb
djgutier: We discussed this briefly in the meeting today, but just to document here with some more points.  Reasoning for having as a helper is to 
- attempt at paralleling existing code in Expertiza,
- deal with the fact that there isn't a request path to handle (rather a delayed task), and
- allow for a Simicheck comparison to be started from anywhere which makes code changes later easier
Also when the Simicheck server finishes the comparison it notifies the Expertiza server with a request to a callback URL that is provided, and that is handled in PlagiarismCheckerComparisonController.  It seems like how we wrote the controller demonstrates the last point above.  Does this resolve your question?

spec/models/google_doc_fetcher_spec.rb
djgutier: As discussed in our meeting, I changed this to `expect(res).to be_falsely` but the test fails.  Apparently in ruby empty strings are truthy!  
```
  1) WebsiteFetcher fetches a bogus website and returns an empty string
     Failure/Error: expect(res).to be_falsey
       expected: falsey value
            got: ""
     # ./spec/models/website_fetcher_spec.rb:24:in `block (2 levels) in <top (required)>'
```
The interface here was intended to be that fetch_content performs the request, sanitizes the content, and returns a string.  Here is how the model is used
```
        content = fetcher.fetch_content
        if !content.empty?
          self.upload_file(id, team.id, content, file_number)
        end
```
Not sure how you'd like to proceed at this point, please advise.

djgutier:
Updated unit testing details on the wiki: http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Spring_2016/E1738_Integrate_Simicheck_Web_Service#Automated_Testing_within_Rails

PR #952, opened 2017-05-07
E1743-Click and impersonate

apan255:


config/routes.rb
```diff
@@ -6,6 +6,7 @@
   resources :submission_records
   get 'auth/:provider/callback', to: 'auth#google_login'
   get 'auth/failure', to: 'content_pages#view'
+  post 'impersonate/impersonate', to: 'impersonate#impersonate'
```
Winbobob: This line is useless. We already had this route.

Winbobob:
See a631605de8e4062647e859fde17cdce3079ced27

Winbobob:
## Feedback
- Method `can_impersonate?` will slow the page loading time. And each time click `next`, the code will traverse all users. We may find a way to optimize this process.
 - Clicking username to impersonate user is convenient. But in some reports, a new column was added in order to show unityid for the impersonate functionality. We think adding another column is superfluous. The better way may be to add the unityid just before users' full names, like `unityid (full name)`. 
 - We refactored and merged impersonate part into the master code base.

PR #954, opened 2017-05-09
E1732: Additions to logging

kira0992:


yangsong8:
Based on our discussion on May 19th, the logging feature will be done as files in Expertiza server instead of database tables. Each day, Expertiza creates a new file and each month Expertiza creates a new folder. The parsing of the log files can be done outside Expertiza.

PR #955, opened 2017-05-09
Moodle Expertiza Integration

joshio1:
LTI (Learning Tools Interoperability) support added for integrating Expertiza into Moodle.
See http://wiki.expertiza.ncsu.edu/index.php/Moodle_Expertiza_Integration/


app/controllers/lti_assignment_users_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+    attributes["name"] =  attributes["email"].split("@")[0] if attributes["email"] != nil?
+    attributes["fullname"] = params['lis_person_name_given']
+    attributes["email_on_submission"] = 1
+    attributes["email_on_review"] = 1
+    attributes["email_on_review_of_review"] = 1
+    attributes
+  end
+
+  def create_new_user(attributes)
+    user = User.new(attributes)
```
yangsong8: I don't understand this, why this code is in lti_assignment_users_controller.rb?

app/controllers/grades_controller.rb
```diff
@@ -194,11 +194,18 @@ def save_grade_and_comment_for_submission
     @team = participant.team
     @team.grade_for_submission = params[:grade_for_submission]
     @team.comment_for_submission = params[:comment_for_submission]
+    successful = false;
```
yangsong8: This design should be improved, see Dr. Gehringer's comments below:

> Bad design!  The save method should be in charge of pushing to the LMS, not the grades controller.

> Which save method?  The save method that lets the instructor save a grade.

app/controllers/lti_assignment_users_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+          unless ['LTI-1p0', 'LTI-2p0'].include?(lti_version)
+            (redirect_to redirect_url("Invalid lti_version: #{lti_version}")) and return
+          end
+        else
+          (redirect_to redirect_url("Missing lti_version")) and return
+        end
+      end
+    end
+
+    return [200, nil, nil]
```
yangsong8: I dont really understand this method, more comments should be added.
Plus, what do the returned values mean?

app/controllers/lti_assignment_users_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+  def create_new_user(attributes)
+    user = User.new(attributes)
+    user.parent_id = 2 #Set super-admin id as parent for now
+    user.timezonepref = User.find(user.parent_id).timezonepref
+    user.save!
+    user
+  end
+
+  # Authenticates the user who has requested the Expertiza resource.
+  # Since user is coming from an authentic source like Moodle, user is auto-created if does not already exist.
```
efg: Shouldn't this comment be attached to the preceding method?

app/controllers/lti_assignment_users_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+
+    return [200, nil, nil]
+  end
+
+
+  def lti_assignment_user_params
+      params.require(:lti_assignment_user).permit(:user_id, :assignment_id, :participant_id, :lis_result_source_did, :tenant_id, :grade)
+  end
+
+  def normalize_role(roles_string)
```
yangsong8: I guess this is the role defined by LTI instead of Expertiza. If so, there should be some comments to clarify what ‘role’ it is handling. This is potentially confusing considering Expertiza also has the concept of roles, and they are different than the ones defined in LTI.

app/controllers/lti_assignment_users_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+      end
+      error_code = "LTI_USER_NOT_FOUND"
+      message = "User not registered on Expertiza. Please contact Expertiza administrators for registration details"
+      return [user,error_code,message]
+    end
+    return [user,error_code,message]
+  end # def login
+
+  # Checks the authenticity of the tenant, i.e. LMS like moodle, using the request data
+  def pre_process_tenant
```
efg: This method is too long, and should be broken up.

app/controllers/lti_assignment_users_controller.rb
efg: A constant like "200" needs to be defined as a constant and explained in a comment.

app/controllers/lti_registration_wips_controller.rb
```diff
@@ -0,0 +1,103 @@
+class LtiRegistrationWipsController < ApplicationController
```
yangsong8: Not clear what is handled by this controller. Some class comments are necessary.

efg: What does "wip" stand for?

app/controllers/lti_registration_wips_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+    tenant.secret = @registration.final_secret
+    tenant.save
+
+    @registration.tenant_id = tenant.id
+    @registration.save
+
+    redirect_to_registration @registration, disposition
+  end
+
+  def show_reregistration
```
efg: What is the difference between registration and re-registration?

app/helpers/lti_assignment_users_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+      puts response
+    end
+  rescue Exception => e
+    flash[:notice] = "Could not post score to LMS."
+    puts e;
+  end
+
+  def self.post_grades_to_lms assignment_participants, grade
+    assignment_participants.each do |assignment_participant|
+      user_assignment = LtiAssignmentUser.find_by_participant_id assignment_participant.id;
```
yangsong8: You dont need the ";" in ruby

app/decorators/registrations_controller_decorator.rb
```diff
@@ -0,0 +1,10 @@
+class RegistrationsControllerDecorator
```
efg: Decorators are not mentioned in the writeup.  There should be a comment here saying what they are for.

app/models/assignment.rb
```diff
@@ -298,8 +298,15 @@ def is_microtask?
   # user_name - the user account name of the participant to add
   def add_participant(user_name, can_submit, can_review, can_take_quiz)
     user = User.find_by_name(user_name)
+    add_participant_by_user user,can_submit,can_review, can_take_quiz
```
yangsong8: Is this change related to the Expertiza-LMS integration feature?

yangsong8:
This is a write-up created by the creator of this pull request:
`http://wiki.expertiza.ncsu.edu/index.php/Moodle_Expertiza_Integration`

yangsong8:
Hi @joshio1 , I think you have also made some changes on Moodle. I would suggest you upload your version of Moodle somewhere. Thanks.

app/controllers/lti_registration_wips_controller.rb
efg: "WiP" would be a little clearer.

app/controllers/lti_registration_wips_controller.rb
efg: Omkar explains that it was pre-existing code from the sample LTI app that he didn't change, lest he break anything.  In fact, if Expertiza will never need to re-register in any LTS, we can delete that code.

app/controllers/lti_assignment_users_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+      if(user)
+        assignment_id = params['custom_assignment_id'] # assignment_id needs to be passed as a
+        assignment = Assignment.find(assignment_id)
+        if assignment
+          # Assignment number passed in the request is valid.
+          assignment_participant = AssignmentParticipant.find_by_user_id_and_assignment_id(user.id,assignment_id)
+          lis_result_sourcedid = params["lis_result_sourcedid"]; # sourcedid is used to send grades back
+          @tenant.lis_outcome_service_url = params['lis_outcome_service_url'];
+          unless assignment_participant
+            # Assignment_Participant not found in participant table and assignment exists. So add user to participant table.
```
efg: It seems to me that checking for and/or adding assignment_participant should be done in a different method.

app/controllers/lti_assignment_users_controller.rb
efg: Moreover, the code is **wrong**.  The super_admin should not be the parent of the user, because then the instructor will not ever be able to impersonate this user.  Rather, the parent should be the creator of the assignment.  From this point in the code, we can determine what the assignment is, and then can check the creator of the assignment, and set that creator as the parent of this user.

PR #974, opened 2017-06-30
Local DB Calc- improved pull request

Prashanth261993:


yangsong8:
Hi PR, this pull request somehow has the sentiment analysis code. I would suggest you create a new pull request

Winbobob:
This pull request is messy. I close this pull request right now.
- Even we store the data in MySQL, we still need to change the view. Currently, the view fetches data from ORM (Object Relational Mapper).
- I also tried Redis. It is not fast with objects, so it doesn’t save that much time.
- The whole design should be reconsidered, in order to make the report display faster.
- And also, the current design requires changing almost everywhere scores are displayed.

PR #1013, opened 2017-09-28
Score view

saratkavuru:
Minor Issues :
- Sorting in each of the tabs is not working. Should modify the javascript slightly to recognize <div> by ID and not just class.
-Statistics Tab ->  Should combine both the tables in Participant Statistics.


Winbobob:
LGTM

PR #1025, opened 2017-10-21
E1745. [TLD] Refactor response_controller.rb

dipanshagupta:
Tasks completed:
1. Refactored response_controller.rb (Made changes in 4 existing methods, and created 1 method)
2. Added test cases in response_controller_spec.rb (Wrote 23 test cases)


app/controllers/response_controller.rb
```diff
@@ -3,31 +3,37 @@ class ResponseController < ApplicationController
   helper :file
 
   def action_allowed?
-    case params[:action]
-    when 'edit' # If response has been submitted, no further editing allowed
+    response = user_id = nil
+    action = params[:action]
+    if %w(edit delete update view).include?(action)
```
Winbobob: Good refactoring!

app/controllers/response_controller.rb
```diff
@@ -200,49 +205,28 @@ def pending_surveys
...
-          ]
-        end
-      end
-    end
-
-    # Get all the assignment survey deployments for this user
-    if assignment_participants
-      assignment_participants.each do |ap|
-        survey_deployments = AssignmentSurveyDeployment.where(parent_id: ap.parent_id)
+    [CourseParticipant, AssignmentParticipant].each do |item|
```
Winbobob: Good refactoring!

Winbobob:
Hi team,

* I think you did a  good job!
* For your tests, you put all stubs in `before(:each)` block. I don't think it is a good idea.
    - In your case, every test will execute many unnecessary stubs, which may take more time to run tests.
    - It will be better if you put necessary stubs in each test and only put frequently-used code in `before(:each)` block.

Thanks,
Zhewei

app/controllers/response_controller.rb
```diff
@@ -200,49 +205,28 @@ def pending_surveys
...
+      # Get all the participant(course or assignment) entries for this user
+      participant_type = item.where(user_id: session[:user].id)
+      next unless participant_type
+      participant_type.each do |p|
+        survey_deployment_type = participant_type == CourseParticipant ? AssignmentSurveyDeployment : CourseSurveyDeployment
```
Winbobob: I don't think the logic here is correct. `participant_type` is an array of participants, instead of the type of participant. I think this mistake is due to your misleading variable names. And you just refactor code without varifying the correctness.

PR #1026, opened 2017-10-22
E1788. OSS project Maroon: Heatmap fixes

nielarshi:
Assigned a unique id to all tables based on the round number and included a class "scoresTable" for each table and initialised sorting features based on class.

Added only true values in the resource object returned for the TA, ensured that only courses in which he/she is TA of will be sent back to the view.

Description of the fix is at below wiki link :

[http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Fall_2017/E1788_OSS_project_Maroon_Heatmap_fixes](http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Fall_2017/E1788_OSS_project_Maroon_Heatmap_fixes)

### DIFF
--- a/app/controllers/lottery_controller.rb
+++ b/app/controllers/lottery_controller.rb
@@ -31,7 +31,7 @@ class LotteryController < ApplicationController
       # store each summary in a hashmap and use the question as the key
       teams = JSON.parse(response)["teams"]
       create_new_teams_for_bidding_response(teams, assignment)
-      run_intelligent_bid
+      run_intelligent_bid(assignment)
     rescue => err
       flash[:error] = err.message
     end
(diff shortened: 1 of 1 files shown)


spec/factories/factories.rb
```diff
@@ -157,6 +157,42 @@
     can_choose_topic_to_review true
   end
 
+  factory :assignment_mapping, class: Assignment do
```
Winbobob: I think you do not need this factory since there is one factory for assignment already.

nielarshi: We tried using that, but the other factory has a hard fix on the sequence number for few of the other existing tests. 

The issue is due to this:
sequence(:name, 2) {|n| n = 2; "final#{n}" }

So, for few of the tests, csv file present on the file storage to check with the expected result has this (final2) hardcoded and the check is against it. 

For our scenario, we needed more than 2 assignments in our database but due to the existing code, we couldn't generate more than 2 assignments.

We added the new factory as the older one was not fitting our case. Let us know if we have to fix the existing ones as I feel the above code was made to just make certain tests pass but is not fit for new tests which would need more than 2 assignments.

Let us know your thought.

Issue #311, opened 2013-07-11
Prospective team member should only be able to respond to an ad once


efg:
When a team advertises for new members, someone who responds to an ad should only be allowed to respond once.  Currently, after someone responds to an ad, they can respond again an unlimited number of times.

efg:
Then in E1779, partially merged.  Not clear if issue is fixed, though.

ryanhardy002:
This is still an issue in beta as of 1/27/2022. Possible fix is to add an edit feature in join_team_request_controller and change the view to have an edit button rather than a request button if one request as already been submitted.

johnbumgardner:
> This is still an issue in beta as of 1/27/2022. Possible fix is to add an edit feature in join_team_request_controller and change the view to have an edit button rather than a request button if one request as already been submitted.

this is how id approach it. can you implement this fix?

ryanhardy002:
It is also possible to request to join your own team. This should not be the case.

Issue #227, opened 2012-11-26
Join team --> lose topic


efg:
If a student has a topic, then joins a team that is not already signed up for a topic, the intuitive action should be that the team gets the topic that the joining student has reserved.  In fact what happens is that the student who is joining a team inherits the team's topic, which is null.  So, the unsuspecting student has just lost his topic!  It would be better to have a student joining a team take his topic with him, unless the team already has a topic, in which case the system should inquire which topic should be kept.  It's not quite clear how to do that; who should be asked--the existing eam or the joining member?

yangsong8:
I will test it on my local machine. If this is still an issue, we can make an OSS project in 16 fall.

yangsong8:
This is still true in current master. Student A who does not have topic invites student B who is holding a topic. If student B accepts, student B loose the topic and join student A's team.

yangsong8:
E1652, rejected.

efg:
Part of E1779, which was partially merged.  Was this fixed?

johnbumgardner:
This feature is still not present

PR #1028, opened 2017-10-23
E1779:OSS busgandy:Fix teammate advertisement and request to join a team

Suvodeep90:
**Fix Issue #311:** when one respond to an advertisement, (s)he should only be able to respond once. When a team advertises for new members, someone who responds to an ad should only be allowed to respond once. Currently, after someone responds to an ad, they can respond again an unlimited number of times.
**Fixes Done:** When a user responds to an advertisement then a enter is made in the Join team request table with a participant ID which is unique for each user for each assignment. So we are checking if the that particular user has any entry for that assignment with pending or approved status, if so we are not giving the user an option to respond again until the respond to join a team has been declined. 
**Fix Issue #227 :** if user A got a topic and user B got no topic, then A join B’s team, A’s topic be be dropped and A and B end up with a new team with no topic. This issue should be handle carefully because we cannot simply add B to A’s team (imagine, if A has teammate X and B has teammate Y...). One of a potential fix is that, for assignment w/ topics, one cannot not post an ad unless (s)he holds a topic, similarly, one cannot sent invitations unless (s)he holds a topic. 
**Fixes done:** Here we are checking if an assignment has any topic or not, if a assignment don't have any topic then he/she will have option to send out request. If the assignment has topic then we are checking if the user has any topic or not, if user has selected a topic then only he/she will be able to send out a request


app/helpers/student_teams_helper.rb
```diff
@@ -1,2 +1,16 @@
 module StudentTeamsHelper
+  # This method will return true if the assignment have some topic
+  def topic?(assignment_id)
```
Winbobob: This method is duplicated to an [existing one](https://github.com/expertiza/expertiza/blob/master/app/models/assignment.rb#L54).

app/helpers/student_teams_helper.rb
```diff
@@ -1,2 +1,16 @@
...
+    topics = SignUpTopic.where("assignment_id = ?", assignment_id)
+    topics.empty?
+  end
+
+  def user_has_topic?(user_id, assignment_id)
```
Winbobob: It will better to put this method in `team.rb`. Since topics will be assigned to teams.

Suvodeep90:
So do you want me change and push the changes again?

ferryxo:
Yes please. Thanks.

Suvodeep90:
Changed the method location from helper to team.rb

Winbobob:
See #1129

PR #1032, opened 2017-10-24
E1757 - Introduce a Student View for instructors

Aminoid:
OSS project to introduce student view for instructors


app/views/menu_items/_suckerfish.html.erb
```diff
@@ -3,14 +3,27 @@
   item = session[:menu].get_item(item_id)
   selected = session[:menu].selected?(item.id)
   long_name = item.name.split(/\W/).collect{|word| word.capitalize}.join(' ')
+  student_view = session.key?(:student_view) ? ([1, 26, 27, 2, 14].include? item_id) ? true : false : false
```
Winbobob: Hard-coding menu item ids is not a good idea.

Issue #961, opened 2017-06-05
Can't turn off metareview deadline


efg:
If an assignment has a metareview deadline, e.g., because it has been copied from another assignment with a metareview deadline, but you don't want the new assignment to have metareviews, you can go to the Due Dates tab and deselect "Use metareview deadline".  The deadline will appear to go away.  But when you save and refresh the screen, the deadline reappears.  Moreover, in at least one situation, this causes View scores to crash.

efg:
Ferry finds that this bug is fixed.

Issue #972, opened 2017-06-14
When edit an assignment, changing the number of review rounds doesn't work


ferryxo:
Using the code on Github (June 14, 2017), the button to update the number of review rounds doesn't work. It throws Uncaught TypeError: due_dates_table_element_list[i].id.split(...).last is not a function

ferryxo:
also, the 2 popups (confirm before changing and after changing the rounds) are not necessary. 1 is good.

efg:
This has been fixed.

PR #1033, opened 2017-10-24
E1778 OSS project Purple: UI fixes for assignment creation

rajanalwan:
Fixed issues #961 and #972.

By:
Vidhyalakshimi Sreenivasan - vsreeni
Rajan Alwan - ralwan

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,8 @@ and switch to another environment if you don't like the one you started with.
 See the Expertiza wiki for setup instructions. Please update the wiki with corrections or additional helpful information.
 
  * [OSX](http://wiki.expertiza.ncsu.edu/index.php/Development:Setup:OSX)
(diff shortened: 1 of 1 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -118,6 +123,7 @@ def update
     unless params.key?(:assignment_form)
       @assignment = Assignment.find(params[:id])
       @assignment.course_id = params[:course_id]
+      puts "Perry1"
```
Winbobob: Do not write to stdout.

ferryxo: use "logger.debug" instead. and print messages that make sense e.g., @assignment.course_id instead of senseless strings.

rajanalwan:
New pull request for this project can be found [here](https://github.com/expertiza/expertiza/pull/1136).

PR #1037, opened 2017-10-25
E1769. [TLD] Refactor assignment_form.rb 

XinyuGong:
Objectives: 

-   Refactor `add_to_delayed_queue` method 

    -  Write failing tests first

    -  Split into several simpler methods and assign reasonable names

    -  Extract duplicated code into separate methods

-  Rename method and change all other places it is used

    -  Write failing tests first

    -  `is_calibrated → calibrated?`

-  Use `find_by` instead of `where.first`

    -  Write failing tests first

-  Complete the pending tests in `sign_up_sheet_controller_spec.rb`, and write integration tests for newly-created methods. Refactor corresponding methods, and only then finish pending tests.


app/models/assignment_form.rb
```diff
@@ -115,30 +115,38 @@ def add_to_delayed_queue
     duedates = AssignmentDueDate.where(parent_id: @assignment.id)
     duedates.each do |due_date|
       deadline_type = DeadlineType.find(due_date.deadline_type_id).name
-      due_at = due_date.due_at.to_s(:db)
-      Time.parse(due_at)
-      due_at = Time.parse(due_at)
-      mi = find_min_from_now(due_at)
-      diff = mi - due_date.threshold * 60
+      diff, mi = get_diff_from_now(due_date)
```
Winbobob: Variable names in original code is not reasonable, please rename these variables, like `diff`, `mi`.

app/models/assignment_form.rb
```diff
@@ -115,30 +115,38 @@ def add_to_delayed_queue
...
       next unless deadline_type == "team_formation" and @assignment.team_assignment?
-      dj = DelayedJob.enqueue(DelayedMailer.new(@assignment.id, "drop_one_member_topics", due_date.due_at.to_s(:db)),
-                              1, mi.minutes.from_now)
-      change_item_type(dj.id)
+      add_delayed_job(@assignment, "drop_one_member_topics", due_date, mi)
     end
   end
 
+  # get time difference between due_date and now
+  def get_diff_from_now(due_date)
```
Winbobob: Please rename this method. I think your comment is very clear. 
Why not using `get_time_diff_btw_due_date_and_now`? 
Then you could get rid of the comment.

app/models/assignment_form.rb
```diff
@@ -115,30 +115,38 @@ def add_to_delayed_queue
...
+    due_at = due_date.due_at.to_s(:db)
+    Time.parse(due_at)
+    due_at = Time.parse(due_at)
+    mi = find_min_from_now(due_at)
```
Winbobob: Also, please the variable names here.

Winbobob:
Hi team,

* I think you did a really good job!
* You may need to do some modifications based on my comments above.

Thanks,
Zhewei

app/models/assignment_form.rb
WillSuen: Done! Thank a lot

PR #1039, opened 2017-10-25
Bookmark Code Fix

aapandey:
->In this code fix, we have resolved the issue of back button not functioning properly when a student navigates through the bookmarks and clicks on the back button.
->Code fix is also about adding questionnaire and fixing the rubric system that has trouble with bookmark selection and creation of new questionnaires.

### DIFF
--- a/app/views/bookmarks/list.html.erb
+++ b/app/views/bookmarks/list.html.erb
@@ -70,9 +70,7 @@
 
 <br />
 
-<%= link_to 'New bookmark', :action => 'new', :id => @topic.id %> |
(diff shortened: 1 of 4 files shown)


app/views/bookmarks/list.html.erb
```diff
@@ -70,9 +70,7 @@
 
 <br />
 
-<%= link_to 'New bookmark', :action => 'new', :id => @topic.id %> |
-<%= link_to 'Back', :controller=> 'sign_up_sheet', :action=>'list', :assignment_id => @topic.assignment_id %>
```
Winbobob: Could you explain why you remove the `back` link?

PR #1049, opened 2017-10-25
E1770. [TFD] refactor_assignment_participant.rb

terryliu1995:
first try


app/models/assignment_participant.rb
```diff
@@ -1,6 +1,6 @@
 require 'uri'
 require 'yaml'
-
+require 'TFD1770_refactor'
```
Winbobob: According to [this configuration](https://github.com/expertiza/expertiza/blob/master/config/application.rb#L29), maybe you do not need to require lib file.

app/models/assignment_participant.rb
```diff
@@ -53,6 +55,31 @@ def review_score
...
+    # scores[:quiz][:assessments] = quiz_responses
+    # scores[:quiz][:scores] = Answer.compute_quiz_scores(scores[:quiz][:assessments])
+    scores[:total_score] = assignment.compute_total_score(scores)
+    # scores[:total_score] += compute_quiz_scores(scores)
+    # move lots of calculation from view(_participant.html.erb) to model
+    caculate_scores(scores)
+  end
+
+  # methods extracted from scores method:assignment_questionnaires, merge_scores, topic_total_scores, caculate_scores
+  def assignment_questionnaires(questions, scores)
```
Winbobob: It will be better the method name is a verb or an action, instead of a noun.

app/models/assignment_participant.rb
```diff
@@ -53,6 +55,31 @@ def review_score
...
+    # scores[:quiz] = Hash.new
+    # scores[:quiz][:assessments] = quiz_responses
+    # scores[:quiz][:scores] = Answer.compute_quiz_scores(scores[:quiz][:assessments])
+    scores[:total_score] = assignment.compute_total_score(scores)
+    # scores[:total_score] += compute_quiz_scores(scores)
+    # move lots of calculation from view(_participant.html.erb) to model
+    caculate_scores(scores)
+  end
+
+  # methods extracted from scores method:assignment_questionnaires, merge_scores, topic_total_scores, caculate_scores
```
Winbobob: Good comment! You could move this comment before the definition of `scores` method.

app/models/assignment_participant.rb
```diff
@@ -171,34 +182,14 @@ def bookmark_reviews
...
   # if user does not exist, it will be created and added to this assignment
   def self.import(row, _row_header = nil, session, id)
-    raise ArgumentError, "No user id has been specified." if row.empty?
-    user = User.find_by_name(row[0])
-    if user.nil?
-      raise ArgumentError, "The record containing #{row[0]} does not have enough items." if row.length < 4
-      attributes = ImportFileHelper.define_attributes(row)
-      user = ImportFileHelper.create_new_user(attributes, session)
-    end
+    user = AssignmentParticipant.check_info_and_create(row, _row_header = nil, session)
```
Winbobob: Good refactoring!

lib/TFD1770_refactor.rb
```diff
@@ -0,0 +1,28 @@
+module Instance_method
```
Winbobob: - It will be better if you rename this file to a reasonable name, instead of `TFD1770_refactor.rb`.
- Also, you may need to put these two modules into two lib files since they are quite different.
- Also rename the module names to reasonable names, instead of `Instance_method` and `Class_method`.

Winbobob:
Hi team,

* I think you did a really good job!
* You may need to do some modification based on my comments.

Thanks,
Zhewei

lib/TFD1770_refactor.rb
terryliu1995: I will fix them ASAP, Thank you!

app/models/assignment_participant.rb
terryliu1995: I found that I have to add "require". If I delete "require", the test cannot be passed. 
image

lib/TFD1770_refactor.rb
terryliu1995: I have finished all the modifications except deleting "require", because if I delete it, the test will not be passed. Can you explain this why I cannot delete it?

PR #1053, opened 2017-10-25
OSS - E1783 Convolutional Data Extraction From GitHub

kaggarw2:


Gemfile
```diff
@@ -11,6 +11,7 @@ gem 'bind-it'
 gem 'capistrano'
 gem "capistrano-db-tasks", require: false
 gem 'capistrano-rails'
+gem 'chartkick'
```
Winbobob: You could use `highchart` javascript library, which has already been used in Expertiza.

app/views/submission_records/index.html.erb
```diff
@@ -29,4 +29,23 @@
       </tr>
   <% end %>
   </tbody>
+
+  <%
+     if @authors.nil?
+     else
+  %>
+  <h4> Commits Over Project Timeline </h4>
+  <%= line_chart @authors.map{ |goal|
```
Winbobob: Please move the logic from view to helper.

Gemfile
RickKwok: I mentioned using `highchart`

kaggarw2:
Sorry, but the team had no prior knowledge on javascript and we had a time constraint as maximum time was invested in setting up the dev environment.
We had mentioned in our early design that we are keen to use chartkick for graphs. 

Sent from Mail for Windows 10

Gemfile
kaggarw2: Sorry, but the team had no prior knowledge on javascript and we had a time constraint as maximum time was invested in setting up the dev environment.
We had mentioned in our early design that we are keen to use chartkick for graphs.

PR #1055, opened 2017-10-25
Project E1763. Fixing Staggered Deadlines. 

ashah22:
We have made the following changes:
1)  A student is allowed to sign up for topics based solely on the deadline of the topic. The assignment deadline does not matter.
2) The instructor can now duplicate topics. It copies all the original topic fields. The max_choosers is however changed to the available slots of the original topic.
3) Instead of having to edit the deadline for each topic, the instructor is now allowed to select multiple topics from a assignment with staggered deadline and assign a due date to each of them by entering the date just once in the date assignment box.
4) The date assignment box for the selected topics automatically takes the latest deadline of all the topics. Thus, that will be the default the deadline. The instructor does not need to manually fill all the boxes and can only change the date if he so wishes.
5) The multiselect also supports shift+click. If an instructor selects a single topic, then press shift and select the other topic, all the topics in the middle will automatically be selected. Thus the instructor need not select all topic one by one.
6) The topics in all assignments are now sorted based on the topic_identifier when displayed.


Winbobob:
Redo in the 2018 fall semester final project ([E1860](https://github.com/expertiza/expertiza/pull/1314)).

PR #1060, opened 2017-10-25
E1749. [TFD] Refactor questionnaires_controller.rb

shreyaszagade:
Wrote test case for QuestionnairesController#create in questionnaires_controller_spec.rb
Note: One of our teammate was not well so we were given concession from our mentor Zhewei Hu. So we had to do 2/3 of the work.
But we have tried our best to complete as much as we can.


spec/controllers/questionnaires_controller_spec.rb
```diff
@@ -1,4 +1,20 @@
-describe QuestionnairesController do
+extend RSpec::Matchers
```
Winbobob: You do not need to add the code at the top of this test file. Please remove them.
`should` is a deprecated RSpec keyword, you need to use `expect` instead.

spec/controllers/questionnaires_controller_spec.rb
```diff
@@ -1,4 +1,20 @@
...
+
+    # ...or disable the `should` syntax...
+    c.syntax = :expect
+
+    # ...or explicitly enable both
+    c.syntax = [:should, :expect]
+  end
+end
+
+describe QuestionnairesController, type: :controller do
```
Winbobob: You do not need to add `type: :controller` here. RSpec will figure it out automatically.

Winbobob:
Hi team,

- You only refactor one method.
- Your tests are not completed.

I hope you could commit your work ASAP.

Thanks,
Zhewei

app/controllers/questionnaires_controller.rb
```diff
@@ -107,7 +82,7 @@ def create_questionnaire
 
   # Edit a questionnaire
   def edit
-    @questionnaire = Questionnaire.find(params[:id])
+    @questionnaire = Questionnaire.find_by_id(params[:id])
```
Winbobob: Please follow the syntax on Line 268 (`find_by(id: xxx)`).

Issue #696, opened 2016-05-13
Instructors can make changes to each others' rubric


yangsong8:
I impersonated Dr. Ringleb ("sringleb") and I found that she can change Dr. Gehringer's rubric.

efg:
I hope she doesn't!

yangsong8:
According to the discussion on May 13, an instructor can see the public rubric created by other instructors, however, we do not allow the instructor to make changes to the rubric unless the current user is the owner, the TA of the owner or the admins (including super-admins).

efg:
The only change that should be possible to an existing rubric (that has been used to create at least one response) is editing the wording of criteria.  If they want to do anything else, they have to copy the rubric and then edit it.

Also, there is a question about whether a rubric should even be editable after an assignment has finished.  If you can, then in effect, you can "change history."

efg:
Currently, you get a list of all public review rubrics.  If you try to edit one that's not yours, then you get a notifier that you're not allowed to edit it.  But it would obviously be better if the edit icon just didn't appear for rubrics you can't edit.

Also, by default only your rubrics should be shown.  You should have to click a box to "Show public items", which used to be the case for assignments, but now seems to have been removed.

Saurabh110:
Fixed in pull requests #1061

Issue #577, opened 2015-08-17
[project idea] Fix dumping and loading rubric criterion from csv


yangsong8:
This seems to be an OSS project to me.

Expertiza team did a significant refactoring on questionnaire and questions in summer 2015. The function for dumpling and loading questions in questionnaires should be re-written:
- There is no session field in questionnaire table now. We do not treat the ordinary rubrics and custom rubric separately. 
- true_false field in questions table is also gone.
- Dumping and loading question should work for all the existing question types.

You should re-write:
- questionnaire_helper.get_questions_from_csv
- questionnaire_helper.create_questionnaire_csv

yangsong8:
Part of E1563

"Test (and fix if they are broken) the export and import functionality. Write test for importing and exporting questionnaires.  "

yangsong8:
1563 was rejected.

yangsong8:
When you work on this, plz make sure you use the import/export controller.

efg:
There are _lots_ of problems here. 

- There's a long header row that mentions fields (e.g., "crypted password") that have nothing to do with rubrics.
- Exported text is not surrounded by quote marks, which means that if there's a comma in a rubric prompt, the comma will be treated as the end of the field.
- There's an "Export Detail" section of the page that (1) doesn't export anything, and (2) gives a list of fields that couldn't relate to a rubric item (e.g., "Score").  Likely these fields belong to responses.  So maybe it should be, "Export responses".

efg:
This is related to E1936, Specialized rubrics, which was merged.

Specialized rubrics are to Review rubrics what role-based reviewing is to Teammate Review rubrics.  So Role-based reviewing should reuse code from specialized rubrics.

johnbumgardner:
closed now that role based reviewing is in Expertiza

PR #1061, opened 2017-10-26
E1765	[OSS project Brown: was E1715, should be updated] Rubrics

jithinjohn7:
Fixes for 
Issue #696 
Issue #577

### DIFF
--- a/app/models/question.rb
+++ b/app/models/question.rb
@@ -87,4 +87,62 @@ class Question < ActiveRecord::Base
     end
     question_ids
   end
+
+  def self.import(row, _row_header, session, _id = nil)
+    if row.length != 5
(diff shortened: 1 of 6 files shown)


app/models/question.rb
```diff
@@ -87,4 +87,62 @@ def self.get_all_questions_with_comments_available(assignment_id)
...
+      raise ArgumentError, "Questionnaire Not Found"
+    elsif qid > 0
+      question = Question.find_by_id(qid)
+      attributes = {}
+      attributes["txt"] = row[0].strip
+      attributes["type"] = row[1].strip
+      attributes["seq"] = row[2].strip.to_f
+      attributes["size"] = row[3].strip
+      attributes["break_before"] = row[4].strip
+      question.questionnaire_id=_id
```
Winbobob: Should `_id`  be `id'?

jithinjohn7: the variable used is correct. 
Is the query about naming convention?

Issue #404, opened 2014-02-21
Assigning to a course needs to move submission directory!


efg:
When an assignment is assigned to a course, and it already has submissions, those submissions need to be moved to a place under the directory for the course.

Here's why.  When an assignment is created outside of a course, the instructor specifies a path for the directory where the submissions are supposed to be kept.  This path begins with the instructor's directory.  But, when an assignment is created inside a course, the path begins with the course directory … which is a subdirectory of the instructor's directory.  So when an assignment is assigned to a course, its submission directory normally needs to be moved.

HOWEVER … even when creating an assignment outside of a course, the instructor may have specified a submission directory that has the course name in the pathname, e.g., csc517/s14/assignment1.  So, before moving the submission directory, the code needs to check that the course directory is not already an initial substring of the submission directory.  If it is, the submission directory should not be moved.

Winbobob:
If the instructor try to move an assignment from one course to another course, the directory needs to be moved too.

efg:
This was part of E1782, OSS Project Red, which is planned to be merged.

Saurabh110:
The project E1782 is merged in the Master branch #1070

Issue #391, opened 2013-12-07
Prevent assignments from sharing a directory


efg:
We have had many bad situations when two assignments share a submission directory.  Usually this has been from the bug in copying assignments (which I hope is still fixed).  Today it happened because an instructor (karl.schmitt) created two different assignments without specifying a submission directory, so that both of the assignments were in his home directory.

To prevent this, the system should warn when ..
- creating an assignment but not specifying a submission directory (though this may be perfectly reasonable if the instructor only expects links to be submitted), and
- creating an assignment that uses a submission directory that is already in use by another assignment.

yangsong8:
Might be fixed, I will double check

efg:
Also see the (long) discussion in #253

yangsong8:
I have checked this, this issue is till in Expertiza.

yangsong8:
This is still in Expertiza.

efg:
This is still extant as of June 2018, though it is addressed by E1782, Assignment directories, which is planned to be merged.

PR #1070, opened 2017-10-28
E1782 Project Red - Issues #391 & #404 fixed

daxamin:


spec/models/assignment_spec.rb
```diff
@@ -316,4 +316,31 @@ def generated_csv(t_assignment, t_options)
...
+
+  #describe '#check_directory_path' do
+  #  create(:assignment)
+  #  @assignment = create(:assignment)
+  #  expect(check_directory_path).to eq(false)
+  #end
+
+  describe 'check_directory_path' do
+    it 'returns false if directory path already in use' do
+      @assignment = create(:assignment)
```
Winbobob: For unit test, I think you do not need to create objects. You could `build` objects instead.

PR #1071, opened 2017-10-28
E1746. [TFD] Refactor user.rb

tfei3:
1.write tests in user_spec.rb.
2.refactor methods in user.rb


app/models/user.rb
```diff
@@ -320,20 +296,15 @@ def is_teaching_assistant_for?(student)
...
 
   def self.search_users(role, user_id, letter, search_by)
-    if search_by == '1' # search by user name
-      search_filter = '%' + letter + '%'
-      users = User.order('name').where("(role_id in (?) or id = ?) and name like ?", role.get_available_roles, user_id, search_filter)
-    elsif search_by == '2' # search by full name
-      search_filter = '%' + letter + '%'
-      users = User.order('name').where("(role_id in (?) or id = ?) and fullname like ?", role.get_available_roles, user_id, search_filter)
-    elsif search_by == '3' # search by email
+    key_word = {'1'=>'name','2'=>'fullname','3'=>'email'}
```
Winbobob: Very good refactor!

spec/models/user_spec.rb
```diff
@@ -1,3 +1,4 @@
+include Rails.application.routes.url_helpers
```
Winbobob: I am not sure whether you need this line or not.

config/environments/test.rb
```diff
@@ -34,3 +34,5 @@
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
 end
+
+Rails.application.routes.default_url_options[:host]='localhost:3000'
```
Winbobob: I am not sure whether you need this line or not.

Winbobob:
Hi team,

* I think you did a really good job.
* In order to make TravisCI pass, you have to rename `config/database.yml`  back to `config/database.yml.example`. And the correct way to create `database.yml` is copy `database.yml.example` and remove `.example`.

Thanks,
Zhewei

spec/models/user_spec.rb
haooline: I added this line in order to use url_for method in user_rspec.rb, i tried to remove this line and Rails.application.routes.default_url_options[:host]='localhost:3000', but I get a NoMethodError.

Winbobob: hmm, I think you need this line. Very good point!
Later I would move this line to RSpec config file.

app/models/instructor.rb
```diff
@@ -32,4 +32,20 @@ def self.get_my_tas(instructor_id)
     end
     ta_ids
   end
+
+  def self.get_user_list
+    participants = []
+    Course.where(instructor_id: self.id).find_each do |course|
```
Winbobob: It seems that you just simply move the code, without check the functionality.
Your refactor does not work!

tfei3: Sorry about that, we will fix that as soon as possible.

PR #1090, opened 2017-11-03
Add a case for topic_suggestion_spec and optimize review_mapping_spec

nathan-du:
It would cover anonymous suggestion.


nathan-du:
It should work when instructor approves anonymous suggestion topic. However, since we didn't store id for anonymous suggestion topic, expertiza would return an error. That's why travis failed.

PR #1102, opened 2017-12-01
E1791. Track the time that students look at the other submissions --> logging improvement

jerry-shijieli:
Accomplishment by our team:
1. Refactor link review time record codes from previous team.
2. Build database and add new controller to record review time.
3. Use jQuery and embedded ruby code to open/download links and files on review page.
4. Use Ajax to send review data to record link/file review time.


app/views/response/_submitted_files.html.erb
```diff
@@ -6,4 +6,73 @@
     <div id="files" style="display:none">
       <%= display_directory_tree(participant, @files, true) %>
     </div>
+    <script type="text/javascript">
```
ferryxo: maybe extract this as a separate JS file

Winbobob:
See #1124

PR #1103, opened 2017-12-01
E17A4. Allow calibration to be part of an assignment

Harika123m:
In order to gain practice, the Student should be able to give calibration reviews before providing peer reviews. Thus, the Instructor should be able to enable calibration for every new and existing assignment. By this way, the student can improve peer reviewing by comparing his/her results with expert reviews for the same assignment.


Winbobob:
What is the project id of your project?

app/views/student_review/_responses.html.erb
```diff
@@ -65,7 +65,8 @@
                               <%= link_to "Edit", {:controller => 'response', :action => 'edit', :id => @latest_response.id} %>
                           <%end%>
                         </td>
-                      <% else %>
+                      <!-- Update link is disabled for all calibrated reviews -->
```
Winbobob: This part looks weird to me. Why are update links disabled for calibrated reviews?

valli03: Update link is disabled for calibrated review because calibration review happens only in one round unlike the peer reviews which take place in two or more rounds. So after submitting the review for calibration, the student will not be able to update the review and since there is no further round of calibration, we disabled the update link.

Winbobob:
We will try to merge this pull request after confirming the design with the professor.

PR #1108, opened 2017-12-02
E17A6. [TDD] Fix account creation over web to work reasonably

terryliu1995:
1. Users can add a new institution when they want to request an account.
2. When a user is requesting a new account, using wrong information should see the flash message on the view.
3.  Add an option in “Administration > Show…” menu
4. After super-admin or admin approves the request, make sure the record does not disappear on the page. There should be an email sent to email address offered by the requester. 
5. Make the email address clickable.


app/controllers/users_controller.rb
```diff
@@ -217,8 +219,17 @@ def request_user_create
     # TODO: Do not allow duplicates
     # TODO: All fields should be entered
     @user = RequestedUser.new(user_params)
-    @user.institution_id = params[:user][:institution_id]
-    @user.status = 'Under Review'
+    if params[:user][:institution_id] == ""
```
Winbobob: I don't think the logic here is correct.

Winbobob:
Partially merged, see d949b7335536b3bdb5b6dc28beb8b62ee5bdf213.

PR #1126, opened 2017-12-06
E1797 - Timestamp for student submissions

ShyamKatta:
introducing Visualization for students to view their submissions of each assignment.


app/views/student_task/view.html.erb
```diff
@@ -1,104 +1,353 @@
+<script src="https://cdnjs.cloudflare.com/ajax/libs/vis/4.21.0/vis.min.js"></script>
+<link href="https://cdnjs.cloudflare.com/ajax/libs/vis/4.21.0/vis.min.css" rel="stylesheet" type="text/css" />
+<style type="text/css">
+  body, html, input {
```
ferryxo: should be done as a separate css file

app/views/student_task/view.html.erb
```diff
@@ -1,104 +1,353 @@
...
+     if due.deadline_type_id.eql? 1
+       { :id => due.id, :start=> due.due_at, :className => "submissionDue", :content => "Round "+(due.round.to_s)+"<split>Submission due by "+'<br>'+due.due_at.strftime("%m/%d/%Y at %I:%M %p") }
+     else
+       { :id => due.id, :start=> due.due_at, :className => "reviewDue", :content => "Round "+(due.round.to_s)+"<split>Review due by "+'<br>'+due.due_at.strftime("%m/%d/%Y at %I:%M %p") }
+     end
+   end
+%>
+
+<!-- display only if submissions are made-->
+<% unless @team.nil?
```
ferryxo: maybe put most of the conditional logic in the controller

app/views/student_task/view.html.erb
```diff
@@ -1,104 +1,353 @@
...
+       else
+         @href_arr.push(submission.content)
+         { :id => submission.id, :start=> submission.created_at, :className=> "hyperlinkUpload" ,:content => submission.content+'<split>'+submission.created_at.strftime("%m/%d/%Y at %I:%M %p") }
+       end
+     end
+
+   end %>
+
+<!-- Reviews not yet started -->
+<% unless @review_mappings.nil?
```
ferryxo: same problem here. i think there are too much conditional checking in the view that could be done in the controller. we should minimize the logic in the view since it's mixing with html.

PR #1144, opened 2018-03-07
E1804. OSS project Yellow: Topic management

dburcal:


app/assets/stylesheets/signup.css.erb
```diff
@@ -0,0 +1,63 @@
+.jsgrid-grid-body {
+
+  overflow-x: auto;
+  overflow-y: auto;
+  -webkit-overflow-scrolling: touch;
+
+}
+
+.jsgrid-table {
+  width: 150%;
```
ferryxo: this could be the reason why it goes beyond the page width

PR #1148, opened 2018-03-18
Logging capability

nielarshi:
Add Expertiza specific logger to log different severity logs in
different files.
Capture actions of users into logs.


nielarshi:
@Winbobob : Could you please have a look at this. I have fixed code climate issues. There is a conflict with a recent change.

PR #1156, opened 2018-03-26
E1812. Write unit tests for on_the_fly_calc.rb

ashwinrisbood:
First commit


ysgeer1:
Thank you for the feedback.
Climate issues are fixed. Still 2 remaining, not sure how to fix.
Issue 1 - fixed: round_sym, grades_by_rounds and total_num_of_assessments are removed from lines where the code can assign them the values.
Issue 2: we never get that issue when we run tests, but the added line should fix the problem.

Winbobob:
Thank you for your quick response! The remaining 2 code climate issues are fine. I will look into your tests.

app/models/on_the_fly_calc.rb
```diff
@@ -60,11 +61,11 @@ def scores(questions)
         calculate_assessment
       else
         assessments = ReviewResponseMap.get_assessments_for(team)
-        score_team[:scores] = Answer.compute_scores(assessments, questions[:review])
+        score_team[:scores] = Answer.compute_scores(assessments, questions)
```
Winbobob: I dont think this modification is correct. Please check the `compute_scores` method in `Answer` model.

Winbobob:
With test skeletons.

PR #1158, opened 2018-03-27
E1806. OSS Project Juniper: Bookmark enhancements

ShrinathCheriyana:


app/models/questionnaire.rb
```diff
@@ -27,7 +27,8 @@ class Questionnaire < ActiveRecord::Base
                          'GlobalSurveyQuestionnaire',
                          'Course SurveyQuestionnaire',
                          'CourseSurveyQuestionnaire',
-                         'BookmarkratingQuestionnaire',
+                         'Bookmark RatingQuestionnaire',
```
Winbobob: Below is the questionnaire types in the current database:
```
'ReviewQuestionnaire'
'MetareviewQuestionnaire'
'AuthorFeedbackQuestionnaire'
'AssignmentSurveyQuestionnaire'
'CourseSurveyQuestionnaire'
'TeammateReviewQuestionnaire'
'GlobalSurveyQuestionnaire'
'QuizQuestionnaire'
```
Since all questionnaire types do not have white space in them.  
- Can you get rid of the questionnaire types with white spaces? 
- Also, you can remove line 33, which is my outdated comments.

PR #1159, opened 2018-03-28
E1800. Add past-due assignments to task list

vishymakthal:
The documentation for our changes can be found [on our wiki](http://wiki.expertiza.ncsu.edu/index.php/CSC/ECE_517_Spring_2018_-_E1800:_Add_past-due_assignments_to_task_list). 

Our tasks included:

- Highlight with specific colors current and next due dates. 
- Check for correction in due dates of assignments
- Possibly think about a way to separate the list of teammates from the task list. Currently, it lists teammates that the student has worked with anytime during the semester, right in the task list. It would make sense to list teammates separately, somewhere else on the page. You can separate them, if needed share your ideas/brainstorm with me.
- Show required action on: every student has to review others work and give feedback. 
- Write all the required tests before implementing/refactoring the methods in the above mentioned classes.


Winbobob:
Testing objects empty or not is very "shallow". These tests did not consider the values or the attributes of objects.

PR #1168, opened 2018-04-09
E1812. Write unit tests for on_the_fly_calc.rb

Winbobob:


Winbobob:
### General comments
- When I tried to run your tests, they did not work. After skipping 4 tests, the rest of tests can pass. But the test coverage is only 13.01%. Basically, you did not write any complete tests.
- Please fix code climate issues as many as you can
### Issues
- Many of your tests do not have expectations.
- Some of your expectations are useless, eg.,
```
allow(Assignment).to receive(:varying_rubrics_by_round?).and_return(true)
expect(assignment2.varying_rubrics_by_round?).to eq(true)
```
You expect the behavior you just mocked, which is totally meaningless.

PR #1174, opened 2018-04-27
E1821 - Migration Commit

kalyanghosh:
E1821 - Regularization of Expertiza Database

Proposed Changes:-
 - Created a rails migration for deadline_types table to remove duplicate entry (Signup,  ID = 9).
 - Updated records in due_dates tables which referenced deadline_types table with ID = 9.

@Winbobob


Winbobob:
See #1777

PR #1178, opened 2018-04-29
E1817: Adding Student Generated Questions to Rubric

ShrinathCheriyana:
Please put the title of your pull request in the following format: *Project ID. Name*
for example, `[Project id/Independent study/Issue fix/...]. project name`, eg., `E904. Integrating courses and assignments on Expertiza with Moodle`

We suggest you include below information in your pull request:

### DIFF
--- a/app/controllers/questionnaires_controller.rb
+++ b/app/controllers/questionnaires_controller.rb
@@ -363,6 +363,31 @@ class QuestionnairesController < ApplicationController
     valid
   end
 
+  # Methods for Supplementary Review Questionnaire
+
+  def create_srq
+    @participant = AssignmentParticipant.find(params[:id])
+    @team = Team.find(@participant.team.id)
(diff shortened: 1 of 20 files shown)


app/controllers/questionnaires_controller.rb
```diff
@@ -363,6 +363,31 @@ def valid_quiz
     valid
   end
 
+  # Methods for Supplementary Review Questionnaire
+
+  def create_srq
+    @participant = AssignmentParticipant.find(params[:id])
+    @team = Team.find(@participant.team.id)
+    if @team.srq_id.nil? then
+      @questionnaire = Questionnaire.new
```
Winbobob: Questionnaire creation code can use `create` instead. And it is unnecessary to use the instance variable. Also, you can make values as parameters, instead of assigning them one by one.

PR #1219, opened 2018-10-14
E1853. Write unit tests for menu.rb 

jgiallo:
Initial Pull Request for E1853 so that TAs can track progress. Do Not Pull This In Yet.


filegit.txt
```diff
@@ -0,0 +1,7 @@
+{\rtf1\ansi\ansicpg1252\cocoartf1404\cocoasubrtf470
```
Winbobob: I do not think you need to commit this file :).

PR #1248, opened 2018-10-25
E1826 : Add student view for instructors

hrshagrwl:
## E1826 : Add Student View for Instructors

Create a student view for instructors so that they can perform student operations without impersonating them. When an instructor logs in, he/she is in the default view with corresponding menu items (with the exception of the 'Assignments' menu item). There should be an option to switch to and revert back from the student view. When in student view, an instructor is presented with the student menu as listed above.

https://www.youtube.com/watch?v=Hwr0czX0qpI

mentor : @efg

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,7 @@ Setup
 
 ### NCSU VCL image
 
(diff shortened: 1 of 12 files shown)


akanksha0625:
The staggered deadlines test case passes in our local system. Please check if there has been any changes made to the the Travis test cases.

PR #1253, opened 2018-10-27
E1834. Improve e-mail notifications

pratik-abhyankar:
*E1834. Improve email notifications*

- Changes
1. Added email functionality to assignment_participant.rb file when user is added by importing CSV file.
2. Also added email functionality to course_participant.rb when user is added by importing CSV file.

- Mentor
Dr. Edward Gehringer


app/models/assignment.rb
```diff
@@ -365,6 +365,13 @@ def stage_deadline(topic_id = nil)
     due_date.nil? || due_date == 'Finished' ? due_date : due_date.due_at.to_s
   end
 
+  # E1834 Fall 18
+  def review_deadline(topic_id = nil, num_review_rounds)
```
efg: Not a good idea to put a method like this in the assignment class.  Is assignment going to have methods for all due-date types?  This seriously damages the cohesion of assignment.rb, which is a complicated class anyway.

app/models/assignment_participant.rb
```diff
@@ -287,4 +290,16 @@ def stage_deadline
     end
     stage
   end
+
+  # E1834 Fall 18
+  def is_round_valid_for_mail?
+    topic_id = SignedUpTeam.topic_id(self.parent_id, self.user_id)
+    return assignment.review_deadline(topic_id, assignment.num_review_rounds)
```
efg: A method whose name ends in "?" should be returning a boolean.

app/models/assignment_participant.rb
```diff
@@ -287,4 +290,16 @@ def stage_deadline
...
+  end
+
+  # #E1834 Fall 18
```
efg: Remove commented-out code.

app/models/course_participant.rb
```diff
@@ -27,6 +27,9 @@ def self.import(row_hash, _row_header = nil, session, id)
     unless CourseParticipant.exists?(user_id: user.id, parent_id: id)
       CourseParticipant.create(user_id: user.id, parent_id: id)
     end
+    password = "password"#user.password
+    prepared_mail = MailerHelper.send_mail_to_user(user, "Your Expertiza account and password have been created.", "user_welcome", password)
```
efg: This duplicates code from above.  Make it a call to another method, which is implemented in one place.

app/controllers/submitted_content_controller.rb
```diff
@@ -36,7 +36,8 @@ def view
   end
 
   def submit_hyperlink
-    @participant = AssignmentParticipant.find(params[:id])
+    # @participant = AssignmentParticipant.find(params[:id])
+    participant = AssignmentParticipant.find(params[:id])
     return unless current_user_id?(@participant.user_id)
```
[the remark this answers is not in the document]
pratik-abhyankar: We have corrected the code in commit id e09761b7d8746f26485f264c7fd7936b8d956326. The code refers to @participant at all places in the method now.

app/controllers/submitted_content_controller.rb
```diff
@@ -57,6 +58,23 @@ def submit_hyperlink
       end
       ExpertizaLogger.info LoggerMessage.new(controller_name, @participant.name, 'The link has been successfully submitted.', request)
       undo_link("The link has been successfully submitted.")
+
+      # E1834 Fall 18
+      # Send email to reviewers to review new submission, if review_round is valid and not last.
+      if participant.is_round_valid_for_mail?
```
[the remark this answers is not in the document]
pratik-abhyankar: We have refactored the name of this method as per the suggestion in commit id e09761b7d8746f26485f264c7fd7936b8d956326

app/controllers/submitted_content_controller.rb
```diff
@@ -57,6 +58,23 @@ def submit_hyperlink
...
+        #reviewers = []
+        participant.reviewers.each do |reviewer|
+          #reviewers << User.find(reviewer.user_id).email
+          review_mappings = ResponseMap.where(reviewer_id: reviewer.id)
+          review_mappings.each do |map|
+            reviewed_object_id = ''
+            if map.reviewee_id == @participant.id || map.reviewee_id == @participant.team.id
```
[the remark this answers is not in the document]
pratik-abhyankar: On further inspection, we found that the id passed in the URL suffix was incorrect. We referred to reviewed_object_id instead of lastest_response.id. 
We have corrected it now and the new code changes make it much better. 
Commit id: e09761b7d8746f26485f264c7fd7936b8d956326

app/models/assignment.rb
pratik-abhyankar: We have refactored the code and the method is removed from assignment class.

app/models/assignment_participant.rb
pratik-abhyankar: Yes, the method returns a boolean value.

app/models/assignment_participant.rb
pratik-abhyankar: We have removed the commented code as per suggestion.

PR #1295, opened 2018-11-16
Bawagne2 master

ajcpendyala:
Merging Project E1815.1. 

- Removed the sentiment analysis (color coded comments) made by E1820.

- Fixed merge conflicts from E1815.1


ajaypendyala:
@Winbobob Ok I'll do that. Could you tell point me to one place where highcharts are used in Expertiza as an example?

Winbobob:
> highcharts

Hi @ajcpendyala, I did not find the `highcharts` code in Expertiza. I think I am wrong.
Right now we still use `googlecharts` gem. You could:
1. remove the `gchartrb` from the Gemfile;
2. Examples: [grades_controller.rb](https://github.com/expertiza/expertiza/blob/master/app/controllers/grades_controller.rb#L53) and [_team_charts.html.erb](https://github.com/expertiza/expertiza/blob/master/app/views/grades/_team_charts.html.erb#L12).

HTH,
Zhewei

PR #1299, opened 2018-11-27
New docker images and docker compose for Blue-Green deployment

rohilshah95:
The previous Dockerfile was not working, had inconsistency and did not create a new app for every commit. So I created a new docker image from scratch using the stock ruby docker image.

I have modified the docker-compose file to add the blue green docker images. I have also removed the volume and working directory part of the docker-compose since it is not required as the Dockerfile handles that. 

@Winbobob @efg


Winbobob:
LGTM

PR #1307, opened 2018-11-30
E1868. Remove reports from review_mapping_controller.rb

RamyaViju:
1. Adding code from E1837
2. Creating new reports_controller and reports view
3. Moving all the views corresponding to the reports to the new reports view
4. Modifying routes.rb to include the new changes
5. Deleting the unused reports_helper.rb file
6. Modifying the UI button from view review report -> view reports
7. Modifying the mappings to point the right controller method


RamyaViju:
Assignment Branch Condition issues cannot be resolved. This issue says that there are too many assignments within a block of code but these assignments are to populate the instance variables and render them to the views.
If we try resolving this issue, we will be stuck with another issue stating that "the line is too long". Hence we will not be resolving these issues.

PR #1317, opened 2018-12-05
Adding Load Balancer file and reorganizing folders.

siddhantshah:
Adding Load Balancer file and reorganizing folders.
@efg 
@Winbobob


Winbobob:
LGTM

PR #1334, opened 2018-12-13
Independent study - DevOps in Expertiza

Winbobob:


Winbobob:
Since Expertiza does not have that many users now, containerizing Expertiza by using docker will add another layer between the server and the application, which will make the application slower. Therefore, we decide to use the second server as a testing server and use `Capistrano` to handle the deployment.

PR #1367, opened 2019-03-09
E1914. Refactor users_controller.rb

nikitaparanjape:
Refactoring the method get_role:
1) The method does not change any record in the database. It also includes some selection logic. Hence it makes sense to keep it in the controller.
2) The method is now renamed to 'role'. 'get_role' is more of a Java way of naming methods. 'role' sounded more ruby-ish. 

Documentation for methods show and show_selection:
1) Written comments for the key differences between the two methods.
2) Written comments for what the each of the methods do.
3) Written comments for what each block inside the method does.

Documentation for the 'foreign' method in which specifies:
1) What the method as a whole does.
2) What each line does.
3) What the method is being used for.
4) Where the method is being used.

@akshayravichandran


app/controllers/users_controller.rb
```diff
@@ -67,28 +67,48 @@ def list_pending_requested
     @roles = Role.all
   end
 
+
+  # Differences between show _selection and show:
+  #   The key difference between show_selection and the show methods is in how they determine whether the current user has the authority to delete/edit a selected user
+  #   The show method only checks if the current user is not a student or him/herself. It assumes that all other roles are authorised to edit/delete the information of all users.
+  #   The show_selection method allows access only if the current user is higher up the hierarchy of roles than the user s/he is requesting to edit.
+  # The show method additionally calculates the number of total users and the number of groups the user is participating in.
+  # These numbers are used in the show view to display appropriate warning messages when deleting a user.
```
akshayravichandran: You probably don't have to indicate these points (lines 71-76) as the "key differences", as they can be inferred from the code. Instead, it might be more useful to point out the locations from where the `show_selection` method is **called**, versus the locations from where the `show` method is called. This way, we will be able to figure out why these two methods were created and structured the way they are and perhaps be able to refactor them if necessary.

app/controllers/users_controller.rb
```diff
@@ -67,28 +67,48 @@ def list_pending_requested
...
+
+  # determines if the current user is authorised to see/edit the information of the user in params.
+  # the test used to determine is whether the current user is higher up the hierarchy of roles than the user s/he is requesting to edit.
+  # If these permissions check out, the user is redirected to the 'show' view
+  # If these checks come out negative, the user to given an error message and redirected to a list of all users.
   def show_selection
     @user = User.find_by(name: params[:user][:name])
+    #make sure that the requested user exists
```
akshayravichandran: I don't think one line comments like this (and on line 91, 95, 109 and 285) would be necessary. The code beneath this comment is very straightforward. 

However, comments like the ones on line 87 and 288 are useful as it is not immediately clear what the lines of code beneath them do.

app/controllers/users_controller.rb
```diff
@@ -67,28 +67,48 @@ def list_pending_requested
...
       end
+      #redirect to the list of users if the user does not exist.
     else
       flash[:note] = params[:user][:name] + ' does not exist.'
       redirect_to action: 'list'
     end
   end
 
+
+  #finds out the current user's role. If that is not a a student, permission is granted to edit the information of the requested user.
```
akshayravichandran: Following up with my comment on lines 71-76, it may be useful to also indicate where this method is called from (in order to highlight its differences with the `show_selection` method).

app/views/users/_institutions.html.erb
```diff
@@ -1,17 +1,21 @@
...
-    if(this.value == "") {
-      $('#institution_name').show();
-      $('#name_of_other_institution').show();
-    }else{
-      $('#institution_name').hide();
-      $('#name_of_other_institution').hide();
-    }
-  });
-});
+    $(document).ready(function(){
```
akshayravichandran: Were these white space changes (lines 2-12) made because CodeClimate was complaining? If not, then I don't think these are necessary.

app/views/users/_name.html.erb
```diff
@@ -1,2 +1,4 @@
-<tr><td><%= label_tag('user[fullname]', 'Full name (last, first[ middle]): ') %></td><td>
-  <%= text_field 'user', 'fullname', class: "form-control width-250"%></td></tr>
+<tr>
+  <td><label class="label_account_request">Full name (last, first [middle]): </label></td>
+  <td><%= text_field 'user', 'fullname', class: "form-control width-250"%></td>
+</tr>
```
akshayravichandran: Please add a new line at the end of this file for consistent formatting (please also do this at the end of any other file that's missing this)

app/views/users/_self_introduction.html.erb
```diff
@@ -1,2 +1,2 @@
-<p><label for="user_email">Self Introduction</label><br/>
+<p><label class="label_account_request">Self-Introduction: (Please include a link to your website, etc.)</label><br/>
```
akshayravichandran: It might be better to place "(Please include a link to your website, etc.)" as a hint, instead of making it a a part of the label

db/schema.rb
```diff
@@ -715,14 +715,14 @@
   add_index "teams_users", ["user_id"], name: "fk_teams_users", using: :btree
 
   create_table "track_notifications", force: :cascade do |t|
+    t.integer  "notification_id", limit: 4
```
akshayravichandran: I don't think these changes to schema.rb are relevant to your project. Please revert them.

Gemfile
```diff
@@ -30,7 +30,7 @@ gem 'foreman'
 gem 'gchartrb', require: 'google_chart'
 gem 'gdata', require: false
 gem "googlecharts", require: "gchart"
-gem 'haml-rails'
+# gem 'haml-rails'
```
akshayravichandran: Why was this change necessary?

app/models/account_request.rb
```diff
@@ -0,0 +1,2 @@
+class AccountRequest < ActiveRecord::Base
```
akshayravichandran: Shouldn't this model replace the RequestedUser model? I think the validations need to be moved over from that file

app/assets/stylesheets/application.scss
```diff
@@ -26,3 +26,4 @@ $icon-font-path: "bootstrap-sass/assets/fonts/bootstrap/";
 @import "confirm_box";
 @import "inner-pages";
 @import "homepage";
+@import "request_account";
```
akshayravichandran: I think the CSS file must also be renamed to `account_request` for consistency.

app/controllers/users_controller.rb
```diff
@@ -128,7 +129,8 @@ def new
   def request_new
     flash[:warn] = "If you are a student, please contact your teaching staff to get your Expertiza ID."
     @user = User.new
-    @rolename = Role.find_by(name: params[:role])
+    @role = Role.new
```
akshayravichandran: Any reason for creating a new empty `Role`?

app/controllers/users_controller.rb
kunalnarangtheone: Will be fixed in the next update

app/controllers/users_controller.rb
kunalnarangtheone: This has been fixed for 91 and 95

app/controllers/users_controller.rb
kunalnarangtheone: This has been fixed

app/views/users/_institutions.html.erb
kunalnarangtheone: I have no idea where these lines came from and I do not remember putting them here.

app/views/users/_name.html.erb
kunalnarangtheone: I will fix this in the next update

app/views/users/_self_introduction.html.erb
kunalnarangtheone: Fixed

db/schema.rb
kunalnarangtheone: Hi Akshay, I do not remember adding this line, I'm not very sure why it is showing up. Could you please delete this on our behalf. I do not want to mess up anything.

Gemfile
kunalnarangtheone: So, this gemfile generates haml view files in the scaffold, which I commented out while generating the scaffold. I have added it back now in the latest commit.

app/models/account_request.rb
kunalnarangtheone: We have made the required change in the latest commit as per your suggestion. Please take a look and let us if it is fine.

app/assets/stylesheets/application.scss
kunalnarangtheone: So, we got a stylesheet named `account_requests.scss` when we generated the scaffold. We have added the formatting to this file and removed the `account_request`

app/controllers/users_controller.rb
kunalnarangtheone: Do you suggest that we should remove this?

app/controllers/users_controller.rb
nikitaparanjape: All have been fixed.

app/controllers/users_controller.rb
kunalnarangtheone: We have removed this

PR #1394, opened 2019-03-25
E1908. Refactor sign_up_sheet_controller.rb

SrikarKanduri:
Fixed the following issues in sign_up_sheet_controller:

- Create method has an if-else condition determining if create or update should be called. Create method should not be responsible for calling update. Rectified this method by removing call to update and flashing an error instead.
- Update method has a plethora of instance variables defined before updating. These are not necessary (For e.g., look at update method of bookmarks_controller). Refactored the variables not needed out.
- Destroy has a misleading else flash message.  Fixed it.
- Several method names are renamed to be more intuitive. load_add_signup_topics is renamed to get_assignment_data and ad_info is renamed to get_ad.
- Added comments to signup_as_instructor and signup_as_instructor_action methods.
- The list method is too long and is sparsely commented. Provided comments.
- Refactored participants variable to 'teams' in  load_add_signup_topics.
- Signup_as_instructor_action has if-else ladder. It has been made more elegant using a helper function.
- Delete_signup and delete_signup_as_instructor have much in common and violates the DRY principle. Refactored them by moving the duplicate code to a helper function.

### DIFF
--- a/app/controllers/sign_up_sheet_controller.rb
+++ b/app/controllers/sign_up_sheet_controller.rb
@@ -144,6 +144,7 @@ class SignUpSheetController < ApplicationController
     redirect_to controller: 'assignments', action: 'edit', id: assignment_id
   end
 
+  # function to list all topics and bids to a participant
   def list
     @participant = AssignmentParticipant.find(params[:id].to_i)
     @assignment = @participant.assignment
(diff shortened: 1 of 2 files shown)


Winbobob:
Hi team,

Please [sync your branch to the Expertiza `beta` branch](https://help.github.com/en/articles/syncing-a-fork) and make `Travis CI` and `Code Climate` run on your pull request.

Thanks,
Zhewei

PR #1459, opened 2019-04-26
E1927:  Let course staff as well as students do reviews (REAL Beta Branch Pull Request)

mblewallncsu:
This project adds the following enhancements:
-Instructors can do reviews just like a student by using the Add review button on the View_submissions or View_scores pages.
-A special icon has been added to the View_scores page for students to indicate an instructor review.
-A special icon was added to the review page that shows an instructor or TA review.
-Scores from instructor reviews are not calculated in the overall average
-A few test cases have been included.


mblewallncsu:
Threw in the towel on the answer_spec.rb test file.  The problem is that if the course staff submits a response, the total score does not change.  However, if a student submits a response, the total score will change.  This is the way the professor wants the scores to be calculated.  There should be a way of adding a test based on the user's role_id or profile, but there is no documentation on how to add this parameter to the test file.

PR #1538, opened 2019-10-21
E1944. Refactor review_mapping_controller.rb

yashwanthsoodini:
### Proposed changes: 

- student_review_num sounds like a number or an id associated with a student review, while it actually stores the number of reviews that a student can perform. So it is renamed num_reviews_per_student.

- submission_review_num sounds like a number or an id associated with a submission review, while it actually stores the total number of reviews that can be performed on a single submission. So it is renamed num_reviews_per_submission.

- calibrated_artifacts_num sounds like a number or an id associated with the calibrated artifacts, while it actually stores the number of calibrated artifacts. So it is renamed num_calibrated_artifacts. Similarly, uncalibrated_artifacts_num is renamed num_uncalibrated_artifacts.

- participants_hash is not an appropriate name for a hash whose keys are participant ids and values are number of reviews performed by corresponding participants. So it is renamed num_reviews_by_participant_hash.

- add_calibration is a method that changes the attribute of a ReviewResponseMap and has little to do with Review Mapping. So it is now put in a separate controller named ReviewResponseMapController.

- The name add_reviewer may lead the reader to think that the method adds a reviewer to a collection of reviewers (e.g. a list of reviewers). Changing the name to assign_reviewer_manually informs the reader that the method assigns a reviewer (to a submission) and hence improves the readability of the code.

@RamyaViju


VivekReddy98:
> Proposed changes:
> 
> peer_review_strategy is already a complex method in review_mapping_controller.rb, with huge chunks (if-else blocks) dedicated to one single functionality. One such operation is to find a rand_participant id and for which there are 30 dedicated lines in the peer_review_method. This would make the method hard to understand in the already sparsely commented scenario. So these lines were taken into a separate method namely gen_random_participant_id.
> 
> @RamyaViju

VivekReddy98:
> Changes Made on Commit Bug Fixes check_num_reviews_args commit :
> 
> The Check made on num_reviews_per_student and num_reviews_per_submission at the beginning of automatic_review_mapping method is being shifted into a new method namely check_num_reviews_args. This makes the code more readable 
> 
> @RamyaViju

Saurabh110:
Removed switch statement in action_allowed? :

Defined new method 'choose_case(params[:actions])' which takes the name of action as a parameter and returns true or none based on the condition.

This is a more object-oriented approach
@RamyaViju

RamyaViju:
Looks good. Good work guys!

PR #1543, opened 2019-10-22
E1956, There is no shortcut to get free review points: Review Assignment Bug

dhruvil009:
Made Changes to review mapping controller to enforce review limit in the controller extending over the initial version which only disabled it from UI.
Things which are not handled yet: No Flash message to notify user they have exceeded limit for reviews (Done to handle the Rspec testing suite)

@Suraj-Siddharudh


dhruvil009:
Mistake in base branch selection

PR #1544, opened 2019-10-22
E1949. Write Unit Tests for Importing assignment participants and import glitches

dppss91102:
Reason for build failing:
One of our tasks is to add a test for a function that does not exist yet, namely to rename the existing team when there is import conflict. Therefore, as per test-driven development practice, the test will always fail until the function is correctly implemented.


spec/models/team_spec.rb
```diff
@@ -233,18 +293,43 @@
...
       end
+
+      context 'when handle_dups option is insert' do
+        it 'does nothing and returns nil' do
+          expect(Team.handle_duplicate(team, 'no name', 1, 'insert', CourseTeam.new)).to be nil
+        end
+      end
+
+      # By the time this test is added (by E1949), the renaming existing team function does not exist yet,
+      # so it should fail unless the function is implemented and the existing team is renamed and saved.
```
dppss91102: The Travis CI checks failed because it was supposed to fail. Check the commented lines.

RamyaViju:
Hey guys please do work on ensuring that the build passes.

dppss91102:
Hi Ramya,

As discussed in our meetings, one of our tasks is to add a test for a function that does not exist yet, namely to rename the existing team when there is import conflict. Therefore, as per test-driven development practice, the test will always fail until the function is correctly implemented. What do you suggest us to do, delete the test maybe?

Larry

dppss91102:
@RamyaViju Is there anything else that we need to do?

RamyaViju:
Please do mention this in the wiki doc and also in the PR description.

PR #1560, opened 2019-10-27
E1941. Issues related to topic deadlines

theSaurabhMhatre:
We've made the following changes so far:

Created a field to allow drop topic deadlines to be added for individual topics
Handled save/update of drop topic deadlines to DB

### DIFF
--- a/app/views/users/list.html.erb
+++ b/app/views/users/list.html.erb
@@ -34,18 +34,6 @@
     <% end -%>
   </table>
 </div>
-<hr/>
-
-<hr/>
-<%= render :partial => 'shared_scripts/pages' %>
-
-<p>
-  <%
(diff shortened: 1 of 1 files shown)


app/helpers/deadline_helper.rb
```diff
@@ -8,6 +8,9 @@ module DeadlineHelper
   DEADLINE_TYPE_SIGN_UP = 7
   DEADLINE_TYPE_TEAM_FORMATION = 8
 
+  TOPIC_DEADLINE_TYPE = 'TopicDueDate'
+  ASSNT_DEADLINE_TYPE = 'AssignmentDueDate'
```
bharatb4444: Please change ASSNT_DEADLINE_TYPE  to ASSIGNMENT_DEADLINE_TYPE. It is easier to read and understand since this is a constant variable and name should fully describe what is being stored in that variable.

PR #1562, opened 2019-10-28
E1964. Export review scores for projects

pururaj1908:
Adding new functionality to "review_report" view that can export review scores into CSV format and generate downloadable CSV file which can be uploaded into webassign platform.

@hrshagrwl - Please review the changes.


yash97:
This Failure is not related to our addition as you can see above commit named minor changes passed the test but the we made changes in readme.md that failed and even after reverting updation of readme.md test case failed.

PR #1547, opened 2019-10-23
E1958. Two issues related to assignment management

Anmoldesai1997:
Please put the title of your pull request in the following format: *Project ID. Name*
for example, `[Project id/Independent study/Issue fix/...]. project name`, eg., `E904. Integrating courses and assignments on Expertiza with Moodle`

We suggest you include below information in your pull request:
- A description of the changes proposed in the pull request.
- @mentions of the person or team responsible for reviewing proposed changes.

PR #1570, opened 2019-10-28
E1958. Two issues related to assignment management, Issue 1430  #1547

Anmoldesai1997:
E1958. Two issues related to assignment management, Issue 1430  #1547


Anmoldesai1997:
@Yashad

Issue #717, opened 2016-09-01
Account info should be e-mailed to students the first time they are added to an assgt.


efg:
When students' accounts are created by importing a CSV on the Users page, they receive e-mails with their user-ID and password.  But if an account is created by adding them as participants to an assignment when they don't already have an account, e-mail is not sent.  Students should receive e-mails upon account creation, regardless of how their account is created.

yangsong8:
Part of E1655 in fall 2016, rejected.

Issue #345, opened 2013-10-14
System should not tell students to revise last-round reviews


efg:
Evidently, if a submission is revised after review, the system e-mails the reviewer saying to revise the review.  This is just fine ... except if the last round of review has been completed.

The message telling reviewers to revise their reviews should not be sent after the last review deadline has passed.

yangsong8:
E1655Part of E1655 in fall 2016, rejected.

efg:
Project was rejected, so still extant.

Issue #111, opened 2011-09-24
Include links in e-mail notifications


efg:
E-mail notifications from Expertiza say what needs to be done, but don't give the recipient much help in knowing where to turn.
- E-mails about reviews should direct the user to the page where the review is found.
- Deadline reminders should include a link on where to go to perform the needed function.
- Instructor notifications of where reviews disagree by more than a threshold # of points should point the instructor to the reviews that disagree.

The task list collects together some of these page addresses; could they be included in e-mail?  This might require a way to be directed to a particular Expertiza page when one logs in.

This includes issues #33 and #34.

efg:
The third issue (instructor notifications of disagreeing reviews) is fixed, but the other two remain.

PR #1604, opened 2019-11-07
E1940. Improve e-mail notifications

dheerajmakam:
Issue #717: When students' accounts are created by importing a CSV file on the Users page, they receive e-mails with their user-ID and password. But if an account is created by adding them as participants to an assignment when they don't already have an account, e-mail is not sent. Students should receive e-mails upon account creation, regardless of how their account is created.  So this involves adding a call to the e-mailer … or, perhaps, moving an email call from the Users controller to the point where an account is actually created.  [Make sure emailer is running; if you have trouble, email expertiza-support@ncsu.edu.]
Issue #345: Second, evidently if a submission is revised after review, the system e-mails the reviewer saying to revise the review. This is just fine ... except if the last round of review has been completed. The message telling reviewers to revise their reviews should not be sent after the last review deadline has passed.  It would also be nice to fix the message so it tells which review (Review 1, Review 2, etc.) has been revised, and gives the reviewer a link directly to it.
Issue #111: Deadline reminders should include a link on where to go to perform the needed function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 Expertiza
 =========
-
 [![Build Status](https://travis-ci.org/expertiza/expertiza.svg?branch=master)](https://travis-ci.org/expertiza/expertiza)
 [![Coverage Status](https://coveralls.io/repos/github/expertiza/expertiza/badge.svg?branch=master)](https://coveralls.io/github/expertiza/expertiza?branch=master)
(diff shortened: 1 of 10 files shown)


dheerajmakam:
The following test case is failing which is not at all related to our changes, we have tried raising the pull request again yet there is no success.

40) assignment function review strategy tab sets number of reviews by each student
      # review section not yet completed
      Failure/Error: check 'num_reviews_student'
      
      Capybara::ElementNotFound:
        Unable to find visible checkbox "num_reviews_student" that is not disabled
      # ./spec/features/assignment_creation_spec.rb:645:in `block (3 levels) in <top (required)>'

PR #1687, opened 2020-03-09
E2002. Refactor impersonate_controller.rb

nradhak2:
A description of the changes proposed in the pull request.
 - Refactored code in impersonate_controller.rb with the objective to reduce the code complexity, and the lines of code


Saurabh110:
Impersonating from the banner (title bar common on every page) is not working.
There is no 'params[:user]' and yet it gets referenced in function 'check_if_special_char'

PR #1690, opened 2020-03-17
E2012. Refactor lottery_controller.rb

carlklier:
Added a separate methods to handle construct priority info, remove empty teams, create sorted bidding matrix summary, construct team bidding info, and assign available slots so as to reduce method complexity and fix code climate issues.

This PR is a work in progress and still needs review by our mentor @TheAkbar

 Teammates: @uiz-11 @thwinter2


app/controllers/lottery_controller.rb
```diff
@@ -44,20 +35,45 @@ def run_intelligent_assignment
 
   private
 
+  def construct_priority_info(sign_up_topics, teams)
+    priority_info = []
+    # Exclude any teams already signed up
+    teams_not_signing_up = teams.select{|team| SignedUpTeam.where(team_id: team.id, is_waitlisted: 0).any?}
```
carlklier: Won't this include signed up teams rather than skip them?

app/controllers/lottery_controller.rb
```diff
@@ -44,20 +35,45 @@ def run_intelligent_assignment
...
+    teams_not_signing_up.each do |team|
+      # Grab student id and list of bids
+      bids = [0]
```
carlklier: I believe starting the bids with 0 will affect the matrix in the merge_bids_from_different_previous_teams method.

app/controllers/lottery_controller.rb
uiz-11: You are right. Sorry for mishandling this part. Should change select to reject I think?

app/controllers/lottery_controller.rb
uiz-11: Could you illustrate more on this part? Because what I thought was that if bids array contains only [0] then priority_info will not be updated?

app/controllers/lottery_controller.rb
```diff
@@ -44,20 +35,45 @@ def run_intelligent_assignment
...
+      sign_up_topics.each do |topic|
+        bid_record = Bid.find_by(team_id: team.id, topic_id: topic.id)
+        # use safe navigation operator to check that both bid_record and bid_record.priority exist
+        bids << bid_record.priority if bid_record&.priority
```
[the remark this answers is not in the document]
uiz-11: It seems like Ruby 2.2 parser can't recognize the safe navigation operator I used here that's available on version 2.3.

app/controllers/lottery_controller.rb
```diff
@@ -12,64 +12,103 @@ def action_allowed?
...
+  def log(message)
+    ExpertizaLogger.info LoggerMessage.new(controller_name, session[:user].name, message, request)
+  end
+
+  # Generate user bidding infomation hash based on students who haven't signed up yet
+  # Structure of user_bidding_info variable: [{user_id1, bids_1}, {user_id2, bids_2}]
+  def construct_user_bidding_info(sign_up_topics, teams)
+    user_bidding_info = []
+    # Exclude any teams already signed up
+    teams_not_signing_up = teams.reject {|team| SignedUpTeam.where(team_id: team.id, is_waitlisted: 0).any? }
```
carlklier: I think a better name for this variable would be teams_not_signed_up.

app/controllers/lottery_controller.rb
```diff
@@ -12,64 +12,103 @@ def action_allowed?
...
         # [Future work]: we need to find a better way to merge bids that came from different previous teams
-        merge_bids_from_different_previous_teams(assignment, new_team.id, user_ids, priority_info)
+        merge_bids_from_different_previous_teams(assignment, new_team.id, user_ids, user_bidding_info)
       end
     end
-    # Remove empty teams
+  end
+
+  # Destroy current team_user and team_user node if exists
+  def remove_current_user(assignment_id, user_id)
```
carlklier: I think a better name for this method would be something like remove_user_from_previous_team to clarify that the user isn't getting removed but just disassociated from their old team.

app/controllers/lottery_controller.rb
```diff
@@ -78,21 +117,21 @@ def create_new_teams_for_bidding_response(teams, assignment, priority_info)
     end
   end
 
-  def merge_bids_from_different_previous_teams(assignment, team_id, user_ids, priority_info)
-    # Select data from `priority_info` variable that only related to team members in current team and transpose it.
-    # For example, below matrix shows 4 topics (key) and correponding priorities given by 3 team members (value).
+  def merge_bids_from_different_previous_teams(assignment, team_id, user_ids, user_bidding_info)
```
carlklier: I think we could change the assignment param to just sign_up_topics since we only use that and not other attributes of assignment in this method. Would save that first line of code.

app/controllers/lottery_controller.rb
```diff
@@ -78,21 +117,21 @@ def create_new_teams_for_bidding_response(teams, assignment, priority_info)
...
     #   4: [2, 0, 1]
     # }
-    bidding_matrix = {}
-    priority_info.each do |info|
-      next unless user_ids.include? info[:pid]
-      bids = info[:ranks]
-      assignment.sign_up_topics.each_with_index do |topic, index|
-        bidding_matrix[topic.id] = [] unless bidding_matrix[topic.id]
+    bidding_matrix = Hash.new([])
+    current_team_members_info = user_bidding_info.select {|info| user_ids.include? info[:pid] }
```
carlklier: In what case would an entry in user_bidding_info not include :pid key? Is this line necessary?

app/controllers/lottery_controller.rb
carlklier: If you add a zero to the bids array, and then a bid_record is found and the bid_record.priority is added to the bids array, the indices of the priorities will be off by 1 so that when merge_bids_from_different_previous_teams is called, it might see that zero in the 0th index of the user_bidding_info and think that zero corresponds to the priority for the sign_up_topic at the 0th index.

app/controllers/lottery_controller.rb
```diff
@@ -12,64 +12,103 @@ def action_allowed?
...
+  # Generate user bidding infomation hash based on students who haven't signed up yet
+  # Structure of user_bidding_info variable: [{user_id1, bids_1}, {user_id2, bids_2}]
+  def construct_user_bidding_info(sign_up_topics, teams)
+    user_bidding_info = []
+    # Exclude any teams already signed up
+    teams_not_signing_up = teams.reject {|team| SignedUpTeam.where(team_id: team.id, is_waitlisted: 0).any? }
+    teams_not_signing_up.each do |team|
+      bids = [0]
+      sign_up_topics.each do |topic|
+        bid_record = Bid.find_by(team_id: team.id, topic_id: topic.id)
```
carlklier: Where do these bids come from? How do they get created and associated with a team if the team hasn't signed up for a topic yet and the team doesn't have a list of users in it yet? Can someone clarify?

app/controllers/lottery_controller.rb
```diff
@@ -12,64 +12,103 @@ def action_allowed?
...
+  end
+
+  # Generate team bidding infomation hash based on newly-created teams
+  # Structure of team_bidding_info variable: [{team_id1, bids_1}, {team_id2, bids_2}]
+  def construct_team_bidding_info(unassigned_teams, sign_up_topics)
+    team_bidding_info = []
+    unassigned_teams.each do |team|
+      topic_bids = []
+      sign_up_topics.each do |topic|
+        bid = Bid.find_by(team_id: team.id, topic_id: topic.id)
```
carlklier: Where do these bids get created?

app/controllers/lottery_controller.rb
```diff
@@ -12,64 +12,103 @@ def action_allowed?
...
+      # Grab student id and list of bids
+      team.users.each {|user| user_bidding_info << {pid: user.id, ranks: bids} } if bids.uniq != [0]
+    end
+    user_bidding_info
+  end
+
+  # Generate team bidding infomation hash based on newly-created teams
+  # Structure of team_bidding_info variable: [{team_id1, bids_1}, {team_id2, bids_2}]
+  def construct_team_bidding_info(unassigned_teams, sign_up_topics)
+    team_bidding_info = []
```
carlklier: I would change this variable name to something like bidding_info_for_teams to make it clear that this is bidding info for all unassigned teams rather than just a single team.

carlklier: or maybe teams_bidding_info_for_topics

Winbobob:
@Winbobob

Winbobob:
@expertiza-travisci-bot

PR #1691, opened 2020-03-19
E2003, Refactor Assessment 360 Controller

ramyananth:
Fixed the Course student grade summary page.
Added a function for eliminating the divide by zero error
Added a function for reviewing each student's info based on meta_review and teammate_review
Both functions that need to be refactored check if there are any course participants for the specific course. So this functionality is merged for both the functions.

Teammates: @ramyananth (rananth2) @ushvarma (suppala)


app/controllers/assessment360_controller.rb
```diff
@@ -15,21 +15,19 @@ def all_students_all_reviews
     course = Course.find(params[:course_id])
     @assignments = course.assignments.reject(&:is_calibrated).reject {|a| a.participants.empty? }
     @course_participants = course.get_participants
-    if @course_participants.empty?
-      flash[:error] = "There is no course participant in course #{course.name}"
-      redirect_to(:back)
-    end
+    inspect_course_participants(@course_participants)
```
efg: It is not clear from the name of this method what it does.  How about, "insure_existence_of"?

app/controllers/assessment360_controller.rb
```diff
@@ -15,21 +15,19 @@ def all_students_all_reviews
...
     # hashes for view
     @meta_review = {}
     @teammate_review = {}
     @teamed_count = {}
     # for course
     # eg. @overall_teammate_review_grades = {assgt_id1: 100, assgt_id2: 178, ...}
     # @overall_teammate_review_count = {assgt_id1: 1, assgt_id2: 2, ...}
-    %w[teammate meta].each do |type|
+    %w[teammate meta].each do |type|
```
efg: This is a whitespace change, which should be removed.

app/controllers/assessment360_controller.rb
```diff
@@ -15,21 +15,19 @@ def all_students_all_reviews
...
     @teamed_count = {}
     # for course
     # eg. @overall_teammate_review_grades = {assgt_id1: 100, assgt_id2: 178, ...}
     # @overall_teammate_review_count = {assgt_id1: 1, assgt_id2: 2, ...}
-    %w[teammate meta].each do |type|
+    %w[teammate meta].each do |type| 
       instance_variable_set("@overall_#{type}_review_grades", {})
       instance_variable_set("@overall_#{type}_review_count", {})
     end
+
```
efg: Ditto.  Whitespace change with no apparent impact to the code.  Don't confuse readers with empty changes.

app/controllers/assessment360_controller.rb
```diff
@@ -59,21 +57,27 @@ def all_students_all_reviews
...
       # calculate average grade for each student on all assignments in this course
-      if @teammate_review_info_per_stu[1] > 0
-        temp_avg_grade = @teammate_review_info_per_stu[0] * 1.0 / @teammate_review_info_per_stu[1]
-        @teammate_review[cp.id][:avg_grade_for_assgt] = temp_avg_grade.round.to_s + '%'
-      end
-      if @meta_review_info_per_stu[1] > 0
-        temp_avg_grade = @meta_review_info_per_stu[0] * 1.0 / @meta_review_info_per_stu[1]
-        @meta_review[cp.id][:avg_grade_for_assgt] = temp_avg_grade.round.to_s + '%'
-      end
+      review_info_per_student(cp,@teammate_review_info_per_stu, @teammate_review)
```
efg: Thank you for making teammate_review, etc., a parameter, but "review_info_per_stu" is still an unacceptable name for a method.  It doesn't give any hint of what it actually refers to.  Moreover, it seems to be an array of at least two elements, and nowhere does it say what these elements are!  This is no way to write code.  If necessary, make review_info_per_stu into a hash instead of an array, but better to find a way to get rid of it entirely, in favor of something whose purpose is apparent from its name.

app/controllers/assessment360_controller.rb
```diff
@@ -59,21 +57,27 @@ def all_students_all_reviews
...
     # avoid divide by zero error
-    @assignments.each do |assignment|
-      temp_count = @overall_teammate_review_count[assignment.id]
-      @overall_teammate_review_count[assignment.id] = 1 if temp_count.nil? or temp_count.zero?
-      temp_count = @overall_meta_review_count[assignment.id]
-      @overall_meta_review_count[assignment.id] = 1 if temp_count.nil? or temp_count.zero?
+    avoid_divide_by_zero_error(@assignments,@overall_teammate_review_count,@overall_meta_review_count)
+  end
+
+  def avoid_divide_by_zero_error (assignments,overall_teammate_review_count,overall_meta_review_count)
```
efg: This is not a good name for this method.  I cannot understand what it is actually doing.  Probably would be better to break it up into a method that could operate on either teammate_review_count or meta_review_count, and would be called once for each.  
    Even better: instead of teammate_review_count (or overall_teammate_review_count), have a teammate_reviews object and use teammate_reviews.count to return the count.  But that may be too much to ask, since there's only one day left.

app/controllers/assessment360_controller.rb
```diff
@@ -59,21 +57,27 @@ def all_students_all_reviews
...
+    assignments.each do |assignment|
+      temp_count = overall_teammate_review_count[assignment.id]
+      overall_review_count_hash = 1 if temp_count.nil? or temp_count.zero?
+      temp_count = overall_meta_review_count[assignment.id]
+      overall_meta_review_count[assignment.id] = 1 if temp_count.nil? or temp_count.zero?
+    end
+  end
+
+
+  def review_info_per_student(cp,review_info_per_stu,review)
```
efg: Even if the data structure is turned into a hash, you would still need a comment describing what it is used for.  I'm sure it's taken you awhile to figure it out.  Don't make the next reader redo all that work.

app/controllers/assessment360_controller.rb
```diff
@@ -90,48 +94,51 @@ def course_student_grade_summary
...
       @topics[cp.id] = {}
       @assignment_grades[cp.id] = {}
       @peer_review_scores[cp.id] = {}
       @final_grades[cp.id] = 0
-
       @assignments.each do |assignment|
         user_id = cp.user_id
         assignment_id = assignment.id
-
         assignment_participant = assignment.participants.find_by(user_id: user_id)
```
efg: Comments needed saying why next two lines are included.

app/controllers/assessment360_controller.rb
```diff
@@ -90,48 +94,51 @@ def course_student_grade_summary
...
-        unless (peer_review_score.nil? || peer_review_score[:review][:scores][:avg].nil?)
+        unless peer_review_score.dig(:review,:scores,:avg).nil?
           @peer_review_scores[cp.id][assignment_id] = peer_review_score[:review][:scores][:avg].round(2)
         end
       end
     end
   end
 
+
+
```
efg: It seems like only 2 of the 3 parameters should be needed.  If you have assignment_id and course participant, you can find user_id.  And vice versa.  The statement at line 105 should be included in the new method to avoid passing a 3rd parameter.

app/controllers/assessment360_controller.rb
```diff
@@ -90,48 +94,51 @@ def course_student_grade_summary
...
+  end
+
+  def inspect_course_participants(course_participants)
+    if course_participants.empty?
+      flash[:error] = "There is no course participant in course #{course.name}"
+      redirect_to(:back)
+    end
+  end
+
   def populate_hash_for_all_students_all_reviews(assignment,
```
efg: This is a bad name too, because "populate_hash" doesn't give any hint of what this is computing, except that the result will be some hash.

app/controllers/assessment360_controller.rb
ramyananth: Professor, what about the Code climate "Trailing white space" issue that comes up?

app/controllers/assessment360_controller.rb
ramyananth: The hash value for a student is calculated based on the teammate_review_count. Having just one function that can be called with teammate or meta variables might want us to do an extra check if the value passed is a teammate variable or a meta variable

PR #1712, opened 2020-03-30
E2012. Refactor lottery_controller.rb

uiz-11:
Added separate methods to handle construct users and teams info, remove empty teams, create sorted bidding matrix summary, and assign available slots so as to reduce method complexity and fix code climate issues.

Mentor: @TheAkbar

Teammates: @thwinter2, @carlklier, @uiz-11


carlklier:
The majority of our discussion and code review can be seen under the conversation of this pull request: https://github.com/expertiza/expertiza/pull/1690

uiz-11:
Our tests did not use many `build` but instead used `create` because the previous implementation refers to database often and the only way to test the code is to construct a lively objects-filled system.

PR #1772, opened 2020-10-03
E2067: Sign up to signup

ianyehwork:
Resolve **E2067** Issue 10: The name sign_up needs to be signup; it is being used as a noun (sign_up would suggest a verb, the action of signing up).


ianyehwork:
This is a duplicate pull request. Please disregard.

PR #1779, opened 2020-10-07
E2064. Refactor reputation_web_service_controller.rb

uahamedncsu:
Expertiza allows student work to be peer-reviewed, since peers can provide more feedback than the instructor can. However, if we want to assure that all students receive competent feedback, or even use peer-assigned grades, we need a way to judge which peer reviewers are most credible. The solution is the reputation system. Reputation systems have been deployed as web services, peer-review researchers will be able to use them to calculate scores on assignments, both past and present (past data can be used to tune the algorithms).
For this project, our team's job is to refactor the file: reputation_web_service_controller.rb. This file is the controller to calculate the reputation scores. A “reputation” measures how close a reviewer’s scores are to other reviewers’ scores. This controller implements the calculation of reputation scores.

@uahamedncsu
@agupta38
@srcapezza

### DIFF
--- a/app/controllers/reputation_web_service_controller.rb
+++ b/app/controllers/reputation_web_service_controller.rb
@@ -46,12 +46,16 @@ class ReputationWebServiceController < ApplicationController
   #   "order by RM.reviewee_id"
   #
   #  result = ActiveRecord::Base.connection.select_all(query)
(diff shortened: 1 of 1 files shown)


uahamedncsu:
Closing PR as i am creating new one directly from forked beta branch.

PR #1783, opened 2020-10-08
E2063. Refactor tree-display.js and tree_display_controller.rb

nlozevs:
The tree-display.js and its tree_display_controller.rb files are designed to allow Expertiza users to view their Assignments, Courses and Questionnaires at one place. This is the primary control page, as well as the home page for instructors on Expertiza which allows them to create, modify, delete and view Assignments.

The primary problem with this is that both the files, due to their bulky and unoptimized methods, slow the rendering of UI on screen. The methods in these files can be studied and refactored to improve the overall performance of this controller and its corresponding UI. Moreover, any obsolete or unused methods can be removed and DRY principle should be implemented. This project mostly revolves around these 2 files, and would involve refactoring JavaScript more than Ruby on Rails.  Knowledge of JavaScript is a prerequisite for this project.

We will be removing all old/unused methods, and preventing a lot of the data duplication that is slowing down the rendering of the webpage. In addition, we hope to implement some form of lazy loading to prevent wasted time loading data for tabs that the user is not even viewing at the time.

team members:
@nlozevs @ltdonova

### DIFF
--- a/app/assets/javascripts/tree_display.jsx
+++ b/app/assets/javascripts/tree_display.jsx
@@ -1,14 +1,17 @@
 jQuery(document).ready(function() {
   // This preloadedImages function is refered from http://jsfiddle.net/slashingweapon/8jAeu/
   // Actually I am not using the values in preloadedImages, but image loading speed is indeed getting faster
-  var preloadedImages = []
-  function preloadImages() {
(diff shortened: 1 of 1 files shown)


Saurabh110:
Closing the PR because it is made on Master

Issue #1412, opened 2019-03-28
TA authorization bug on "Your scores" page


akshayravichandran:
If a person is listed as a TA in one course and as a student in another course, then if they navigate to the _**"Your scores"**_ page of one of the assignments in which they are participating as a student, they can see a TA's view of that page - effectively allowing them to assign their own grade!

nikolaygtitov:
I'm not sure if there is a bug or system behaves as expected. I was unable to reproduce the exact the same issue, but I found different one. I impersonated a student who was a participant in the CSC/ECE 517, Spring 2019 class and has the following content in his Assignment page:

![temp](https://user-images.githubusercontent.com/8548602/62889826-1e7ac700-bd10-11e9-9066-833d729cfb8d.png)

Note that these assignment he has to complete (or completed) as a student for the class.
After adding this user into the list of TAs for CSC/ECE 517, Fall 2018 class and going back to the assignment view, I see the following list:

![temp1](https://user-images.githubusercontent.com/8548602/62890148-d3ad7f00-bd10-11e9-9e2a-2be93fccd470.png)

Now, all assignments listed in his view are for the class that he is a TA. Moreover, all the assignment that he has to perform (or already performed) as a student are gone.
Is this a bug or this is expected behavior of the system?

akshayravichandran:
This is an expected behavior of the system. As a TA, to view the assignments that you participate in as **student** (your first screenshot), you'll have to click on the red **'Assignments'** tab at the top. The view that you see right after you log in as a TA (your second screenshot) is accessible through **'Manage -> Assignments'**. That view is expected to show only the assignments that you participate in as a **TA**.

The bug I have written up is when you log in as a TA, click on the red **'Assignment'** tab, and then click on one of the assignments that you participate in as a student. After this, if you click on **"Your scores"**, you see a TA's view of that page, rather than a student's view. This is probably happening because the permissions of that view must not just check if the user is a TA, but also whether he or she is a TA for that specific course. I can confirm that the bug has still not been fixed - just checked it with my account.

nikolaygtitov:
You're correct!! Thank you very much for replying. I see that user may assign any grade for himself as TA for the work that (s)he must do as student. This is definitely outstanding issue that needs to be fixed. Thank you for finding this bug!

efg:
Still outstanding in master, but fixed in beta by E2057.

Issue #1351, opened 2019-02-13
There should be a limit to the size of files that can be uploaded ...


efg:
so that randoms can no longer upload long videos.

Saurabh110:
Fixed by pull request #1558

PR #1793, opened 2020-10-12
E2057. Time travel Not Allowed! Restrict TAs’ ability to change their own grade

snkumar32:
Issue #1412:
If a person is listed as a TA in one course and as a student in another course, then if they navigate to the "Your scores" page of one of the assignments in which they are participating as a student, they can see a TA's view of that page - effectively allowing them to assign their own grade!


Mikeyjou:
Issue #1351 
A student can upload files with their submission. In some cases, students upload long videos that might not be necessary for the submission.

PR #1832, opened 2020-11-01
E2083 Revision planning tool E2016

SidharthMehta:
CSC/ECE 517 Fall 2020 - E2083 Revision planning tool E2016

What it does: In the first round of Expertiza reviews, we ask reviewers to give authors some guidance on how to improve their work. Then in the second round, reviewers rate how well authors have followed their suggestions. We could carry the interaction one step further if we asked authors to make up a revision plan based on the first-round reviews. That is, authors would say what they were planning to do to improve their work. Then second-round reviewers would assess how well they did it. In essence, this means that authors would be adding criteria to the second-round rubric that applied only to their submission. We are interested in having this implemented and used in a class so that we can study its effect.

@chaitanyamehta
@sidharthmehta
@demadewe
@vickyyang5454

### DIFF
--- a/app/controllers/response_controller.rb
+++ b/app/controllers/response_controller.rb
@@ -316,6 +316,8 @@ class ResponseController < ApplicationController
     @max = @questionnaire.max_question_score
   end
 
+  # gets questions sorted by sequence from review and revision questionnaire,
+  # and merges them in a single list.
   def set_questions_for_new_response
     @questions = sort_questions(@questionnaire.questions)
     if(@assignment && @assignment.is_revision_planning_enabled)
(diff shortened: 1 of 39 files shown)


app/controllers/response_controller.rb
```diff
@@ -312,10 +311,32 @@ def set_content(new_response = false)
     @contributor = @map.contributor
     new_response ? set_questionnaire_for_new_response : set_questionnaire
     set_dropdown_or_scale
-    @questions = sort_questions(@questionnaire.questions)
+    new_response ? set_questions_for_new_response : set_questions  
     @min = @questionnaire.min_question_score
     @max = @questionnaire.max_question_score
   end
+
+  def set_questions_for_new_response
```
chaitanyamehta, line 319: Add comments to function

PR #1847, opened 2020-11-16
E2080. Track the time students look at other submissions

LukeMccon:
We are building on the Previous team's implementation, E1989. Their solution wrote to the database often and did not optimize storage space to the desired metric.

We have implemented a solution by using the built-in function of pstore to mitigate database interaction and reduce the number of entries in the database significantly.

[Wiki](https://expertiza.csc.ncsu.edu/index.php/CSC/ECE_517_Fall_2020_-_E2080._Track_the_time_students_look_at_other_submissions)


Saurabh110:
Decided

Issue #1880, opened 2021-01-20
NoMethodError in Reports#response_report 


johnbumgardner:
<img width="1680" alt="image" src="https://user-images.githubusercontent.com/38589044/105230671-b6025d00-5b33-11eb-8a90-c7154042bbbb.png">

Bug I found trying to look at response_report

johnbumgardner:
Fix was merged into the beta branch

PR #1881, opened 2021-01-20
Issue fix: NoMethodError in Reports#response_report

johnbumgardner:
Three code fixes for #1880: 
Use correct method name to get team color
Verify the existence of SignedUpTeam
Verify the existence of Response

### DIFF
--- a/app/models/on_the_fly_calc.rb
+++ b/app/models/on_the_fly_calc.rb
@@ -51,11 +51,9 @@ public
 
 # Get all of the questions asked during peer review for the given team's work
 def peer_review_questions_for_team(team, round_number = nil)
-  if !team.nil? and !SignedUpTeam.find_by(team_id: team.id).nil?
-    topic_id = SignedUpTeam.find_by(team_id: team.id).topic_id
-    review_questionnaire_id = review_questionnaire_id(round_number, topic_id)
-    Question.where(questionnaire_id: review_questionnaire_id)
-  end
+  Question.where(questionnaire_id: 
+    review_questionnaire_id(round_number, SignedUpTeam.find_by(team_id: team.id).topic_id)) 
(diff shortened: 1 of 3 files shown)


app/models/on_the_fly_calc.rb
```diff
@@ -47,15 +47,21 @@ def compute_avg_and_ranges_hash
...
 def peer_review_questions_for_team(team, round_number = nil)
-  topic_id = SignedUpTeam.find_by(team_id: team.id).topic_id
-  review_questionnaire_id = review_questionnaire_id(round_number, topic_id)
-  Question.where(questionnaire_id: review_questionnaire_id)
+  if !team.nil? and !SignedUpTeam.find_by(team_id: team.id).nil?
+    topic_id = SignedUpTeam.find_by(team_id: team.id).topic_id
+    review_questionnaire_id = review_questionnaire_id(round_number, topic_id)
+    Question.where(questionnaire_id: review_questionnaire_id)
+  else
+    nil
```
Saurabh110, line 59: Ruby methods return 'false' by default, is the else nil part is required?

johnbumgardner: code change was made

app/models/on_the_fly_calc.rb
```diff
@@ -47,15 +47,19 @@ def compute_avg_and_ranges_hash
...
 
-private
+public
 
 # Get all of the questions asked during peer review for the given team's work
 def peer_review_questions_for_team(team, round_number = nil)
-  topic_id = SignedUpTeam.find_by(team_id: team.id).topic_id
-  review_questionnaire_id = review_questionnaire_id(round_number, topic_id)
-  Question.where(questionnaire_id: review_questionnaire_id)
+  if !team.nil? and !SignedUpTeam.find_by(team_id: team.id).nil?
```
Saurabh110, line 54: Use unless to make it more readable

Issue #1882, opened 2021-01-25
Bug in alternate view when looking at scores


johnbumgardner:
Steps to replicate
1. Impersonate student7366
2. Navigate to assignments
3. Navigate to Madeup problem
4. Click alternate view
Get this error screen:
<img width="1680" alt="image" src="https://user-images.githubusercontent.com/38589044/105657866-4d98e000-5e93-11eb-90a0-952da55f4a04.png">

PR #1883, opened 2021-01-25
Issue Fix: Bug in alternate view when looking at scores

johnbumgardner:
Fix for issue #1882. My fixes were:
Ensure that split is never called on a nil object in summary_helper.rb
Make sure that "@summary[round][q]" can be iterated over. There was a problem where this object was resolving to a boolean, and would cause a crash when you try calling each on it. I check to make sure it can be called on

### DIFF
--- a/app/helpers/summary_helper.rb
+++ b/app/helpers/summary_helper.rb
@@ -173,7 +173,7 @@ module SummaryHelper
 
     # convert answers to each question to sentences
     def get_sentences(ans)
-      if !ans.comments.nil? 
+      if !ans.comments.gsub!(/[.?!]/, '\1|')nil? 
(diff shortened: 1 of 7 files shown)


app/helpers/summary_helper.rb
```diff
@@ -173,7 +173,12 @@ def summarize_sentences(comments, summary_ws_url)
 
     # convert answers to each question to sentences
     def get_sentences(ans)
-      sentences = ans.comments.gsub!(/[.?!]/, '\1|').split('|').map!(&:strip) unless ans.comments.nil?
+      if !ans.nil? and !ans.comments.nil?
+        sentences = ans.comments.gsub!(/[.?!]/, '\1|').try(:split, '|') || nil
+        if !sentences.nil?
```
Saurabh110, line 178: Use 'unless' instead of 'if not'

Issue #1884, opened 2021-01-27
Routing Error: No route matches [GET] "/survey_response/view_responses"


johnbumgardner:
Steps to replicate: 
1. Log in as instructor6
2. Go to assignments
3. Find WCAE 2017, click on view Survey Responses
![image](https://user-images.githubusercontent.com/38589044/106010646-acbf4600-6087-11eb-962b-ef3fd197d4aa.png)

PR #1885, opened 2021-01-27
Issue fix: Routing Error for Survey Deployments

johnbumgardner:
When you tried accessing an assignment's survey deployments, it would crash due to a prior refactor from survey_response.rb to survey_deployment_controller.rb. This was documented under #1884 which I recently discovered

I changed files in the UI both under app/views/assignments/edit_other.html.erb and app/assets/javascripts/tree_display.jsx

Also there was another bug where you were looking up the Survey Deployment using the assignment id rather than looking for its parent. 

This may introduce a bug of returning two Survey Deployments if an assignment has two assigned surveys. I'm not sure if this is something that can happen in Expertiza

### DIFF
--- a/app/assets/javascripts/tree_display.jsx
+++ b/app/assets/javascripts/tree_display.jsx
@@ -80,7 +80,7 @@ const node_attributes = {
       }),
       (props) => ({
         title: 'View survey responses',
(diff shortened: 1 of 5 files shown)


johnbumgardner:
@Saurabh110 can you review this one too

Issue #1758, opened 2020-09-17
Refactoring needs in participants_controller.rb


efg:
Refactoring participants_controller was to be an OSS project in Fall 2020, but Aditya did most of the work, and this is all that remains:

Background:  A participant is a user who is participating in a course or an assignment.  The Participant class has two subclasses, CourseParticipant and AssignmentParticipant.  A single controller handles both kinds of participants, as is common for class hierarchies in Rails.  The biggest difficulty with this class is that the code is not very well explained.  It is your job to figure out what it is doing and fix it.  Specifically, …
Issues to be fixed:
* update_authorizations: It is not clear why the method is named as it is, needs a method comment to explain why. 
* Why is there a delete_assignment_participant, but no delete_course_participant?  And should it be just delete?
* Removes a participant from the assignment
* What do has_signature and signature_valid do?  They are set to false, then immediately used in an expression, & not used anywhere else, yet they are local variables!  So why are they even needed?
* Get_user_info does not have spec test case 
* has_signature and signature_valid are essentially not needed
* Unsure if user[:verified] should be equal to permission_granted or false boolean
* They can be removed.

See also https://docs.google.com/document/d/15osKtrilWsLxSdBrp7kbUZC6vi2a9siH8VTgPAUK-qs (the Refactoring minutes).

PR #1888, opened 2021-02-04
Refactoring for ParticipantsController

johnbumgardner:
participants_controller.rb has some things that need fixed according to #1758.

- [x] update_authorizations: It is not clear why the method is named as it is, needs a method comment to explain why.
- [x] Why is there a delete_assignment_participant, but no delete_course_participant? And should it be just delete?
- [x] Removes a participant from the assignment
- [x] What do has_signature and signature_valid do? They are set to false, then immediately used in an expression, & not used anywhere else, yet they are local variables! So why are they even needed?
- [x] Get_user_info does not have spec test case
- [x] has_signature and signature_valid are essentially not needed
- [x] Unsure if user[:verified] should be equal to permission_granted or false boolean
- [x] They can be removed.

Creating PR to run my Spec test and to utilize this PR to organize my work. Not ready for merging.

### DIFF
--- a/app/controllers/participants_controller.rb
+++ b/app/controllers/participants_controller.rb
@@ -50,6 +50,7 @@ class ParticipantsController < ApplicationController
     # E1721 changes End.
   end
 
+  #when you change the duties, changes the permissions based on the new duty you go to
   def update_authorizations
     permissions = Participant.get_permissions(params[:authorization])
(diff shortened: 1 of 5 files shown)


johnbumgardner:
signature_valid and has_signature do nothing

johnbumgardner:
refactoring asks are ready for review @Saurabh110

PR #1905, opened 2021-03-16
E2105. Updated database schema

AvlWx2014:
Adds a column to the `submission_viewing_events` table called `total_time`, which we can use to accumulate total time per link, per round.


AvlWx2014:
This was a mistake.

PR #1910, opened 2021-03-19
E2103. Refactor response_controller.rb

zpparks314:
E2103 Team members:
@zpparks314
@jchinch
@KashML

---
About Expertiza Bot


app/controllers/response_controller.rb
```diff
@@ -101,25 +101,15 @@ def update
   end
```
bahati2000, line 102: Method "new" may need comments at the top

PR #1912, opened 2021-03-20
E2101. Refactor questionnaires_controller.rb

johnHostetter:
Replaced literal values with defined constants to aid in the understanding of code behavior.
Investigated whether the method create_questionnaire is used.
Fixed the methods that have a questionnaire_id as a parameter, as it is not needed, since those methods should be contained within the model class anyways.
Removed one of three checks for QuizQuestionnaire.
Corrected the dropdown's default alternatives to reflect the questionnaire's min and max question score.
Updated the dropdown's RSpec test to reflect the change in behavior.
Discovered more software faults.
Moved three of four private methods from the controller to the model.


Winbobob:
Hi @johnHostetter, please rerun the Jenkins job, then the 2nd warning message will be canceled.

PR #1916, opened 2021-03-25
E2103: Refactor response_controller.rb

zpparks314:
Refactoring the response controller by moving business logic to model, renaming methods, and adding comments.

Team:
@Jchinch
@KashML
@zpparks314


spec/models/response_spec.rb
```diff
@@ -131,6 +131,20 @@
...
 
+  describe '#populate_new_response' do
+    context 'when current round response is found' do
+      it 'returns the current round response' do
+        allow(response).to receive(:populate_new_response).with(:review_response_map, "0").and_return(response)
+        expect(response.id).to eq(1)
+      end
+    end
+    context 'when current round response is found' do
+      it 'returns a new response object' do
```
Winbobob, line 142: This test does not have expectations, therefore, this test will never fail. Please add the expectation for this test.

spec/models/response_spec.rb
```diff
@@ -131,6 +131,20 @@
...
+        allow(response).to receive(:populate_new_response).with(:review_response_map, nil).and_return(:new_response)
+      end
+    end
+  end
+
   describe '.concatenate_all_review_comments' do
     it 'returns concatenated review comments and # of reviews in each round' do
       allow(Assignment).to receive(:find).with(1).and_return(assignment)
```
johnbumgardner, line 150: This test has no expectation as well. Correct this

johnbumgardner (changes requested):

PR #1957, opened 2021-04-29
E2115: Mentor Management

bahati2000:
Please set the title of your pull request in the following format: `[Project id/Independent study/Issue fix/...]. project name`, eg., `E904. Integrating courses and assignments on Expertiza with Moodle`

---
About Expertiza Bot


AvlWx2014:
@johnbumgardner or @Winbobob  - our build is failing due to an error on `rake db:setup`. It says "you have 1 pending migration" and points to the migration file we added in our project. Any insight on this? Is there an update to `schema.rb` or otherwise that we need to push? Perhaps we need to sync with the beta branch again?

app/models/assignment_form.rb
```diff
@@ -2,7 +2,13 @@
...
 class AssignmentForm
-  attr_accessor :assignment, :assignment_questionnaires, :due_dates, :tag_prompt_deployments, :is_conference_assignment
+  attr_accessor :assignment,
+                :assignment_questionnaires,
+                :due_dates,
+                :tag_prompt_deployments,
+                :is_conference_assignment,
+                :auto_assign_mentor
+
   attr_accessor :errors
```
johnbumgardner, line 12: any reason why this is not with the comma separated list above?

app/models/mentor_management.rb
```diff
@@ -0,0 +1,122 @@
...
+    curr_team_size = Team.size(team_id)
+    max_team_members = Assignment.find(assignment_id).max_team_size
+
+    # RuboCop 'use guard clause instead of nested conditionals'
+    # return if the team size hasn't reached > 50% of capacity
+    return if curr_team_size * 2 <= max_team_members
+
+    # RuboCop 'use guard clause instead of nested conditionals'
+    # return if there's already a mentor in place
+    return if team.participants.any? { |participant| participant.duty == Participant::DUTY_MENTOR }
```
johnbumgardner, line 64: why arent you using the def self.user_a_mentor?(user) function here?

app/models/mentor_management.rb
```diff
@@ -0,0 +1,122 @@
...
+    Participant.where(parent_id: assignment_id, duty: Participant::DUTY_MENTOR)
+  end
+
+  # Produces a hash mapping mentor's user_ids to the aggregated
+  # number of teams they're part of, which acts as a proxy for
+  # the number of teams they're mentoring.
+  def self.zip_mentors_with_team_count(assignment_id)
+    mentor_ids = self.mentors_for_assignment(assignment_id).pluck(:user_id)
+
+    return [] if mentor_ids.empty?
```
johnbumgardner, line 113: why return an array when the caller will be expecting a hash. potentially a method could be called on this array object when they expect a hash that could result in a nil:NilClass error

design_document.md
```diff
@@ -6,10 +6,7 @@ This document provides the common design features to be followed while developin
 
 Icons are available in 4 sizes : 16, 24, 32, 48. It is possibility that we might not have all sizes for all icons. Edit the number as per size chose, everything else will remain same.
```
johnbumgardner, line 7: im curious as to how this file was changed in your branch. please revert this back

spec/models/mentor_management_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      team_count = 3
+      r = Random.new(42)
+      team_ids = team_count.times.map {
+        random_id = r.rand(1000..10000)
+        FactoryBot.create(:team, id: random_id)
+        random_id
+      }
+      team_ids.each { |team_id| FactoryBot.create(:team_user, team_id: team_id, user_id: ta.id) }
+      expect(MentorManagement.zip_mentors_with_team_count(assignment.id)).to eq [[mentor.user_id, team_count]]
+    end
```
johnbumgardner, line 121: can you please add a test to make sure your method that sends out the email successfully interfaces with the mailer

johnbumgardner (commented):
Code looks great. Well documented and incredible well designed. I have left a few asks, but for how much is here, this all looks really good.

app/models/mentor_management.rb
AvlWx2014: Fair point! `user_a_mentor?` checks whether there's a corresponding entry in `Participant` for a given `User` with the `mentor` duty. So we could equivalently do:
```ruby
return if team.users.any? &:user_a_mentor?
```

An advantage I can think of for the way it exists currently is that we avoid a bunch of database calls.

app/models/mentor_management.rb
AvlWx2014: The original idea was to return a hash, but `sort_by` returns an Array of Arrays and I forgot to update the method comment. I will fix this.

design_document.md
AvlWx2014: The Markdown table was malformed, and would not render properly here on GitHub. I will still revert the change if you'd like.

app/models/mentor_management.rb
johnbumgardner: sounds good.

design_document.md
johnbumgardner: don't worry about it. i didn't realize there was a problem. good catch!

spec/models/mentor_management_spec.rb
AvlWx2014: We determined such a test is unnecessary since it tests nothing about MentorManagement, just that ActiveMailer works as expected.

app/models/mentor_management.rb
johnbumgardner: i think this looks cleaner.

app/models/assignment_form.rb
AvlWx2014: Not that I know of. It was that way already, but we can move it up there.

spec/models/mentor_management_spec.rb
johnbumgardner: that's fair! we have tests for ActiveMailer elsewhere so this should be okay! Thanks,

app/models/assignment_form.rb
johnbumgardner: yeah combining it would be good.

johnbumgardner:
> @johnbumgardner or @Winbobob - our build is failing due to an error on `rake db:setup`. It says "you have 1 pending migration" and points to the migration file we added in our project. Any insight on this? Is there an update to `schema.rb` or otherwise that we need to push? Perhaps we need to sync with the beta branch again?

most recent build makes it past rake db:setup 👍

app/models/assignment_form.rb
AvlWx2014: Done!

AvlWx2014: This change seems to have caused the Travis build to fail. Not sure how that would come to be.

Issue #2023, opened 2021-08-11
Rubrics Tab Issue


johnbumgardner:
On the Rubrics tab, if a rubric contains no ScoredQuestions, the weight is required to be 0; the system should check for this, and flash a warning.

PR #2034, opened 2021-09-04
Rubrics tab weight check

nnhimes:
PR to check the weight of questions in a rubric and flash a warning if the weight is 0 (has no ScoredQuestions). 
For Johns issue #2023 

Not ready as of 9-4. 
image

Changing the review rubric to my 9-1-21 Test, which one has a section header inside it, does not give my flash warning. May be because the flash[:note] = 'The assignment was successfully saved....' is called immediately after in update_feedback_attributes(). 

Is there a way to flash two messages at once?

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -109,9 +109,6 @@ class AssignmentsController < ApplicationController
     else
       # SAVE button was used (do a redirect)
       redirect_to edit_assignment_path @assignment_form.assignment.id
-      if @assignment_form.rubric_weight_error
-        flash[:note] = "A rubric has no ScoredQuestions, but still has a weight. Please change the weight to 0."
-      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -109,6 +109,9 @@ def update
     else
       # SAVE button was used (do a redirect)
       redirect_to edit_assignment_path @assignment_form.assignment.id
+      if @assignment_form.rubric_weight_error
+        flash[:error] = "A rubric has no ScoredQuestions, but still has a weight. Please change the weight to 0."
```
johnbumgardner, line 113: make this a flash note instead of an error to show simultaneously

app/models/assignment_form.rb
```diff
@@ -106,6 +108,20 @@ def update_assignment_questionnaires(attributes)
...
     attributes.each do |assignment_questionnaire|
+      # Check rubrics to make sure weight is 0 if there are no Scored Questions
+      scored_questionnaire = false
+      questions = Question.where(questionnaire_id: assignment_questionnaire[:questionnaire_id])
+      questions.each do |question|
+        if question.is_a? ScoredQuestion
+          scored_questionnaire = true
+        end
+      end
+      unless scored_questionnaire || assignment_questionnaire[:questionnaire_weight].to_i == 0
```
johnbumgardner, line 119: instead of == 0, use .zero?

johnbumgardner (changes requested):

nnhimes:
Working now, though an unrelated bug was found - adding to issues
image
image

johnbumgardner (approved):

PR #2038, opened 2021-09-12
refactor scores in assignment and participant

johnbumgardner:
grades should not be calculated in assignment or participant, as they are an aggregate of responses, so response map is the best location for these.


nnhimes (approved):
Tested on VirtualBox

- Tree displays
- Assignments can be edited
- Students can be impersonated
- Teams can be created
- Students can submit work
- Students can review each other
- Students can view their scores

nnhimes:
Can also show an assessment360 screen of a courses grades by participant

PR #2039, opened 2021-09-12
Refactor on_the_fly_calc

nnhimes:
Refactoring on_the_fly_calc.rb to be in assignment_helper.rb

_The file has 7 methods: 3 that are exclusively called as assignment instance methods, and 4 that are called internally to help facilitate those 3. The name on_the_fly_calc no longer made sense, and it belongs in helpers_


nnhimes:
Tested on virtualbox - no errors that I could find when editing assignments, creating new ones, or viewing them as a student. 

Expected, since the only real change was the module names and file locations. When methods from the new assignment_score_helper module are called, they are still called on assignment instances. Assignment simply includes the entire module and its methods. Low-risk refactor

app/models/assignment.rb
```diff
@@ -9,7 +9,7 @@ class Assignment < ActiveRecord::Base
   include AssignmentAnalytic
   include ReviewAssignment
   include QuizAssignment
-  include OnTheFlyCalc
+  include AssignmentScoreHelper
```
johnbumgardner, line 12: put these methods into assignment_helper.rb

johnbumgardner (changes requested):

app/models/assignment.rb
nnhimes: New commit should do it

app/helpers/assignment_helper.rb
```diff
@@ -99,4 +99,103 @@ def get_team_name_color_in_list_submission(team)
...
+  # Compute total score for this assignment by summing the scores given on all questionnaires.
+  # Only scores passed in are included in this sum.
+  def compute_total_score(scores)
+    total = 0
+    self.questionnaires.each {|questionnaire| total += questionnaire.get_weighted_score(self, scores) }
+    total
+  end
+
+  def compute_reviews_hash
+    @review_scores = {}
```
johnbumgardner, line 111: why are these being passed as instance variables, can you refactor all of this into using parameters

app/helpers/assignment_helper.rb
```diff
@@ -99,4 +99,103 @@ def get_team_name_color_in_list_submission(team)
...
+  def compute_avg_and_ranges_hash
+    scores = {}
+    contributors = self.contributors # assignment_teams
+    if self.vary_by_round
+      rounds = self.rounds_of_reviews
+      (1..rounds).each do |round|
+        contributors.each do |contributor|
+          questions = peer_review_questions_for_team(contributor, round)
+          assessments = ReviewResponseMap.assessments_for(contributor)
+          assessments = assessments.select {|assessment| assessment.round == round }
```
johnbumgardner, line 132: use select!{ |assessment| assessment.round == round}

app/helpers/assignment_helper.rb
```diff
@@ -99,4 +99,103 @@ def get_team_name_color_in_list_submission(team)
...
+  end
+
 end
+
+private
+
+# Get all of the questions asked during peer review for the given team's work
+def peer_review_questions_for_team(team, round_number = nil)
+  topic_id = SignedUpTeam.find_by(team_id: team.id).topic_id unless team.nil? or SignedUpTeam.find_by(team_id: team.id).nil?
+  review_questionnaire_id = review_questionnaire_id(round_number, topic_id) unless team.nil? or SignedUpTeam.find_by(team_id: team.id).nil?
```
johnbumgardner, line 156: use || instead of or

app/helpers/assignment_helper.rb
```diff
@@ -99,4 +99,103 @@ def get_team_name_color_in_list_submission(team)
...
+    scores
+  end
+
 end
+
+private
+
+# Get all of the questions asked during peer review for the given team's work
+def peer_review_questions_for_team(team, round_number = nil)
+  topic_id = SignedUpTeam.find_by(team_id: team.id).topic_id unless team.nil? or SignedUpTeam.find_by(team_id: team.id).nil?
```
johnbumgardner, line 155: these are literally the same checks so just clean this up

app/helpers/assignment_helper.rb
```diff
@@ -99,4 +99,103 @@ def get_team_name_color_in_list_submission(team)
...
+end
+
+def calc_review_score
+  unless @corresponding_response.empty?
+    @this_review_score_raw = Response.assessment_score(response: @corresponding_response, questions: @questions)
+    if @this_review_score_raw
+      @this_review_score = ((@this_review_score_raw * 100) / 100.0).round if @this_review_score_raw >= 0.0
+    end
+  else
+    @this_review_score = -1.0
```
johnbumgardner, line 167: can we refactor away from using -1.0. This is bad coding practice

johnbumgardner (changes requested):

johnbumgardner (approved):
LGTM

PR #2050, opened 2021-09-26
Assignment helper compute refactor

nnhimes:
We talked about reducing the assignment_helper methods dependence on assignment in our 9-22 meeting. This PR removes all self calls in the methods and allows them to be called outside of assignment now (usually by including assignment as a parameter). 

Did not move these out of assignment_helper because the three main methods to refactor depend heavily on private methods that would need to be copy pasted otherwise. Also, there is no one place that called assignment.method, it was split between assignment_team, response_map, report_formatter_helper, and a handful of spec files.


app/models/response_map.rb
```diff
@@ -126,7 +127,7 @@ def self.participant_scores(participant, questions)
     # Retrieve assignment score
     compute_assignment_score(participant, questions, scores)
     # Compute the Total Score (with question weights factored in)
-    scores[:total_score] = assignment.compute_total_score(scores) 
+    scores[:total_score] = AssignmentHelper.compute_total_score(assignment, scores)
```
johnbumgardner, line 130: can you include assignment helper and not need to actually call it as AssignmentHelper?

app/models/response_map.rb
```diff
@@ -138,7 +139,7 @@ def self.participant_scores(participant, questions)
       scores[:max_pts_available] = topic.micropayment
     end
 
-    scores[:total_score] = assignment.compute_total_score(scores)
+    scores[:total_score] = AssignmentHelper.compute_total_score(assignment, scores)
```
johnbumgardner, line 142: +1

spec/controllers/grades_controller_spec.rb
```diff
@@ -82,7 +82,7 @@
         allow(AssignmentParticipant).to receive(:find).with(1).and_return(participant)
         allow(assignment).to receive(:late_policy_id).and_return(false)
         allow(assignment).to receive(:calculate_penalty).and_return(false)
-        allow(assignment).to receive(:compute_total_score).with(any_args).and_return(100)
+        allow_any_instance_of(AssignmentHelper).to receive(:compute_total_score).with(any_args).and_return(100)
```
johnbumgardner, line 85: not familiar with this syntax.

johnbumgardner (changes requested):
There are quite a few Travis CI failures. Please correct them.

nnhimes:
No obvious fix still, not sure how to get this refactor working - especially when the includes/requires are not working as expected. May need to close the PR and start from scratch

nnhimes:
Making a new fresh attempt, too many commits here

PR #2070, opened 2021-10-04
E2122. Refactor impersonate_controller.rb

duhaoze11:
The current change:
- Add a new line in README.md for test.
- Have finished refactoring.
- Have finished writing tests.

> We sent a message to you about the missing ajax plugin for auto-complete last week, but we still haven't received any response yet. The other branch `duhaoze11/expertiza/2012` is the commit with the proper plugin, which was finished in Oct 2012. This missing plugin caused the failed test. Additionally, an unworkable usage of ruby getting data from the database was found in the files. The details are in the message we sent to you.

- @johnbumgardner @duhaoze11


duhaoze11:
> When we try to test the auto_complete_for_user_name function, it failed as the below image shows: image

This is what I talked about the missing ajax plugin in the first message.

duhaoze11:
And in the 4-th test of the check, it seems that these failures were not related to the model impersonate we are working on.

PR #2081, opened 2021-10-11
E2146. Introduce a Student View for instructors

WeiRui-Wang:
E2146. Introduce a Student View for instructors

+ Implemented navigation bar in student view for instructor.
+ Implemented dynamic menu item retrieval to increase maintainability.


WeiRui-Wang:
NOTE:
During an extended black-box testing performed by E2146 feature team, an **pre-existing issue** was found. Such pre-existing issue impacted the accessibility of "Pending Surveys" option in the Student View mode.
We spend some time on analyzing the potential cause of the issue, after a comprehensive review of the yml, we observed that the role_student.yml working as intended as represented by the URL of "[url: "/response/pending_surveys"](https://github.com/Aeront39/expertiza/blob/6c46d720692f3e7eca457c60bef383fa621957e7/config/role_student.yml#L252)".
With the finding in additional to further code review on all the changes we have made, as we also pull the beta branch directly from `expertiza:beta` and we compared to the URL that used by the live expertiza, we concluded that the issue of inaccessible "Pending Surveys" was caused by **pre-existing commit** that break the internal routing control.

Live Expertiza
image

Beta Main Student View
image

Implemented Instructor Student View
image

WeiRui-Wang:
Hi, @Winbobob. We encountered `OpenSSL::SSL::SSLError:` issue in Travis CI and we wonder if we need to do something in order to fix the issue. Thanks!
https://app.travis-ci.com/github/expertiza/expertiza/jobs/543316107

Aeront39:
Hello! As noted by my teammate WeiRui-Wang, we are unsure how, if at all, the changes we made could have caused the errors listed in the Travis reports.

Job 11926.1 describes an error that occurred in `staggered_deadline_sepc.rb`. When investigating the error listed, however, it seems like the issue was due to a mistake in initialization of the test object, not due to any navigation bar changes. A snapshot of the error is provided below:
image
As can be seen in the highlighted region, an error message is generated saying that the Author Feedback was not provided when initializing Assignment1665, not allowing the topic deletion to occur. Reviewing the spec test seems to confirm that no author feedback was provided with the assignment, and there does not appear to be any obvious way the addition of the "Student View" button could have affected this. I thus would like to request this error be skipped.

Job 11926.4 describes two errors occurring in `/spec/helpers/review_mapping_helper_spec.rb`, both of which contain the error condition `OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed`. As can hopefully be evidenced by the files we edited, our contributions to this feature should not have impacted this in any way. From a cursory search for this error, it seems to typically occur when the ruby installer is missing an SSL certificate. For the sake of providing complete information, this error will also be given as an image below:
image
Investigating the line where this error occurs within the spec file also does not seem to provide the code described in this error. I may be misunderstanding the error report, but assuming I'm not, potentially Travis is checking an older version of this spec file, or something of that sort? Regardless, it seems highly unlikely the additions we made would have caused this test to fail, and so I'd like to request these errors also be skipped.

My team and I believe that the scope of the feature we created was small enough that issues such as these are unlikely to have been caused by our changes. If such is not the case, we would be more than happy to make the changes necessary to correct them. However, at the moment we don't see how these errors could have been caused by the additions we created. I'd like to request these errors be ignored, or if not, potentially some direction toward how the changes we made could have caused these issues, so that we may correct them.

I hope we can continue to collaborate to improve expertiza!

PR #2098, opened 2021-10-17
E2134. Write unit tests for admin_controller.rb

yhslo:
This part is mainly for rspec tests for admin_controller.rb, including action_allowed? authentication functions and all #list and #show functions.
The tests will invent a set of new users, and for list, it will only test the output of instance variables; for show, it will test instance variables and the rendering.

List of the controller tested:
- action_allowed?
- show_super_administrator
- list_super_administrators
- show_administrator
- list_administrators
- list_instructors
- show_instructor


nnhimes (approved):
Lots of tests, nice

PR #2114, opened 2021-10-20
E2117. Adding changes to remove create_questionnaire function

ProfDoof:
All explanation of changes will occur in below comments as each individual makes their changes


ProfDoof:
@isha-bansal0115 feel free to explain your changes here

PR #2155, opened 2021-11-28
E2155. Calibration submissions should be copied along with calibration assignments

Kasimir123:


app/models/assignment_form.rb
```diff
@@ -407,7 +439,16 @@ def self.copy(assignment_id, user)
     old_assign = Assignment.find(assignment_id)
     new_assign = old_assign.dup
     user.set_instructor(new_assign)
-    new_assign.update_attribute('name', 'Copy of ' + new_assign.name)
+    # Set name of new assignment as 'Copy of <old assignment name>'. If it already exists, set it as 'Copy of <old assignment name> (1)'.
+    # Repeated till unique name is found.
+    name_counter = 0
```
johnbumgardner, line 444: each with index

app/models/team.rb
```diff
@@ -291,4 +291,42 @@ def self.remove_user_from_previous_team(parent_id, user_id)
     team_user = TeamsUser.where(user_id: user_id).find { |team_user| team_user.team.parent_id == parent_id }
     team_user.destroy rescue nil
   end
+
+  #Create new teams for calibrated assignments with respect to the old team already present
+  def self.createnewteam(old_assign, new_assign_id)
+    puts "Alpha"
```
johnbumgardner, line 297: delete

johnbumgardner (changes requested):

nnhimes (changes requested):
Can you add comments to the new methods? I'm having trouble following what some things are supposed to do

johnbumgardner:
Im not going to merge this project. There is some unexpected behavior I think ought to be fixed, and we need unit testing on the new methods added, but this is moving in the correct direction.

PR #2157, opened 2021-11-29
E2167 - Testing Team Related Files

MathewAaron:
Please set the title of your pull request in the following format: `[Project id/Independent study/Issue fix/...]. project name`, eg., `E904. Integrating courses and assignments on Expertiza with Moodle`

---
About Expertiza Bot


nnhimes:
Issue comment pasted:

The' update' method fails because the methods called inside 'begin' statements cant be stubbed. The '.save' call on stubbed objects are not successful because stubbed objects can't be saved in the database.

cc: @priyajakhar

nnhimes:
Issue comment pasted:

Among 15 tests the def index method fails due to missing template. I tried to look around for the template so that I could redirect to it but was unnsuccessful.

Rest all the test for this controller passes, with coverage 85.71%

johnbumgardner:
@nnhimes i added an index file. in the controller i added a check to make sure the user had admin privileges to view it. wrote up some corresponding spec tests

johnbumgardner (approved):

db/schema.rb
```diff
@@ -30,8 +30,9 @@
     t.integer  "tag_prompt_deployment_id", limit: 4
     t.integer  "user_id",                  limit: 4
     t.string   "value",                    limit: 255
-    t.datetime "created_at",                           null: false
-    t.datetime "updated_at",                           null: false
+    t.datetime "created_at",                                                    null: false
+    t.datetime "updated_at",                                                    null: false
+    t.decimal  "confidence_level",                     precision: 10, scale: 5
```
nnhimes, line 35: Why is line 35 being added? I can't find another place confidence_level is used in this PR

PR #2161, opened 2021-11-30
E2163. Refactor Waitlist Functionality

asingh072318:


nnhimes:
@johnbumgardner Can you look at this? Says Waitlist table doesn't exist, even though waitlist has existed prior to this project. I didn't have the table in my SQL though. Do we need a migration?

johnbumgardner:
> @johnbumgardner Can you look at this? Says Waitlist table doesn't exist, even though waitlist has existed prior to this project. I didn't have the table in my SQL though. Do we need a migration?

we dont need a waitlist table, this are all class methods use anyway, BUT i do think that there isnt much utility in having a model called waitlist. how challenging would it be for you to refactor to the lib directory?

PR #2173, opened 2022-01-04
added delete_all method for teams_controller

divyang02:
Added delete_all method for teams_controller, for allowing "Delete All" functionality in teams for a course.

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -68,6 +68,15 @@ class TeamsController < ApplicationController
     @team = Team.find(params[:id])
   end
 
+  def delete_all
+    # Deletes all the teams assigned to a course, by finding them via node_object_id
+    root_node = Object.const_get(session[:team_type] + "Node").find_by(node_object_id: params[:id])
+    child_nodes = root_node.get_teams.map { |e| e.node_object_id}
+    course_teams = Team.where(id: child_nodes)
+    course_teams.destroy_all if course_teams
+    redirect_to action: 'list', id: params[:id]
+  end
+
   def delete
     # delete records in team, teams_users, signed_up_teams table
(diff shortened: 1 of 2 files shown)


app/controllers/teams_controller.rb
```diff
@@ -68,6 +68,15 @@ def edit
     @team = Team.find(params[:id])
   end
 
+  def delete_all
+    # Deletes all the teams assigned to a course, by finding them via node_object_id
+    root_node = Object.const_get(session[:team_type] + "Node").find_by(node_object_id: params[:id])
+    child_nodes = root_node.get_teams.map { |e| e.node_object_id}
+    course_teams = Team.where(id: child_nodes)
+    course_teams.destroy_all if course_teams
```
namanshrimali, line 76: I have one small suggestion to make on lines 75-76. Instead of fetching the Team data and then deleting, we can directly delete the data from child_nodes array. I was able to make the following code run achieving same results:
```ruby
Team.delete_all(id:child_nodes) if child_nodes
```
It has one less db call.

divyang02: I think this actually won't be saving any DB calls, as :-
1. Rails does a lazy evaluation of database queries so it won't be evaluated until we call destroy
2. Also, rails cache the query result so if there is the same DB call, it will be using the cached result instead of fetching it again from the database.

If you want to test the number of queries made, you can install a gem called 'db-query-matchers' on your local branch and test the number of queries made in each case. 
Let me know if you think I missed something :)

namanshrimali: Sure! I will install the gem. If it looks good to you, I'll merge it.

divyang02:
@johnbumgardner This should be closed as it's already been merged into beta by [https://github.com/expertiza/expertiza/pull/2176](url) ?

PR #2174, opened 2022-01-04
removed mutant-rspec

divyang02:
Removed gem mutant-rspec for avoiding failing Travis CI build


johnbumgardner (changes requested):
bundle with 1.16.6 please to keep the bundler version consistent with travis

PR #2180, opened 2022-01-13
Response map refactor

johnbumgardner:
refactor on assignment team and response map

PR #2190, opened 2022-01-18
Remove broken report views

nnhimes:
Removal of the two broken report views on the beta branch. These are not in use by the class and were likely added half a decade ago. Also helps prevents problems with future merge #2180.


nnhimes:
All 5 remaining reports view without crashes on the OSS Project & Documentation assignment.

johnbumgardner (approved):
LGTM

PR #2201, opened 2022-01-21
Answer Tagging

johnbumgardner:
Followup to akk5597 pr after fixing merge conflict


nnhimes (approved):
Read through all the non-spec changes, I think its all good/works the same. I haven't used that .select || format before though.

PR #2240, opened 2022-02-08
Improvements to Team Management as an Instructor - Inheritance and Bequeathal

### LINKED ISSUE #2141: Problem with the inherit team button
If you have a course with teams defined at the course level, and then you go to an assignment, manually create teams and assign students.

Every student in my created course is assigned. However, when I clicked the inherit team button, it still gets course teams
image
Students are in two teams at once.

This should either be fixed to disallow inheriting teams when assignment teams are already created, or by deleting all exiting teams and then assigning.

johnbumgardner:
disallow inheriting teams from course when teams exist
refactors bequeathal adds restrictions on the usage of both inheritance and bequeathal. should decrease the number of bugs and crashes in team management

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -119,17 +119,27 @@ class TeamsController < ApplicationController
     redirect_to controller: 'teams', action: 'list', id: assignment.id
   end
 
-  # Copies existing teams from an assignment up to a course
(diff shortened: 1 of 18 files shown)


johnbumgardner:
@nnhimes ready for review and manual testing

nnhimes (approved):
lgtm

Issue #2184, opened 2022-01-16
Viewing Survey Responses on Assignment Crashes


nnhimes:
When editing an assignment, clicking the Etc tab, and then "View survey responses" crashes Expertiza. 

This screenshot is for the OSS Project & Documentation assignment.

![image](https://user-images.githubusercontent.com/41524135/149679205-3a6b3302-e964-40d2-ace8-1c06c2b9fff8.png)

Could be rerouted if no survey deployment found.

johnbumgardner:
I think the best course of action is removing these views for now

nnhimes:
Going to try and reroute it if its simple enough

PR #2241, opened 2022-02-08
Add redirect instead of crash

nnhimes:
Fixes crash from #2184 by redirecting with a warning message

image


johnbumgardner (approved):
LGTM not testing this tho

Issue #2216, opened 2022-01-31
Bugs in deleting a user


efg:
I created a user in error when logged in as efg.  I tried to delete the user, and from my efg account, it told me I didn't have permission.  So I logged in as admin and was able to delete the user; however, the UI gave an error message that it couldn't find the user (a subsequent users listing verified that it had been deleted).

Issue #2233, opened 2022-02-07
Ridiculous loading time on user/list


johnbumgardner:
Even after merging in the bullet gem and eager loading, we still have absurd load times on user/list, mostly because it is retrieving thousands of users and associated record. 

We should refactor the list command to paginate these results, not retrieving all at once.

PR #2245, opened 2022-02-11
Issue fix on the user controller and list,

johnbumgardner:
improvements made in this PR
- no longer load all users on every users/list call
- make sure that the user can only delete another user if they are authorized, as well as offering this option of the users/list page
- letting role be assigned during show.
- minor refactors
Closes issues #2216 and #2233


nnhimes (approved):
Incredible

PR #2248, opened 2022-02-13
grade view bug fixes

johnbumgardner:
Remove buttons to click to survey related 
Remove unnecessary controllers
Remove statistics tab


nnhimes (changes requested):
Can make assignments, teams, and submit work. Clicking view scores as instructor after having two students on different teams give reviews for each other only shows one of the teams' reviews.

image

I gave one team 100% on all questions and the other 60%, so the bar graph at the top works. Going to each team and clicking view scores also shows the correct grades given. So it's just this list that isn't working fully

The review report shows both reviews. No crashes found.

nnhimes:
And delete the Gemfile.lock

Issue #2257, opened 2022-02-17
New users cannot set initial password


ryanhardy002:
When creating a new user, a password is always auto-generated through a welcome email regardless of whether a custom password was entered when creating the user.

ryanhardy002:
Also, attempting to create a user without providing an email address should have an error message that the field cannot be blank. It may be good to set the email options to be checked by default rather than unchecked as well.

PR #2259, opened 2022-02-17
fix issue #2257

ryanhardy002:
Because users are always given a new randomized password, we should remove the field to allow them to specify.

Tested with both instructor creating a new user manually and when you have a account requested. Both cases work with no regressions.


nnhimes (approved):
lgtm if tested

PR #2260, opened 2022-02-19
Auth controller spec

johnbumgardner:
it's called we do a little testing


ryanhardy002 (approved):
LGTM, I see no issues.

Issue #2220, opened 2022-02-01
Default for emails should be the same when creating user from New User form and from importing


efg:
If a user is created from an imported csv, etc., file, then they are set to get email when their work is reviewed, when a new version of work they've reviewed is uploaded, etc.

If a user is created from the user interface, the default is NOT to send them email in these situations.  That should be changed so the behavior is the same in both cases.

johnbumgardner:
Resolved

PR #2261, opened 2022-02-19
Check email pref by default

johnbumgardner:
as mentioned in issue #2220 0, we need to check these preferences by default when creating users to correspond to the default settings when creating a user via import. this PR resolves that problem


ryanhardy002 (approved):
LGTM if manually tested.

PR #2263, opened 2022-02-19
Cop linting WIP 

johnbumgardner:
work slowly on getting linting issues solved


nnhimes:
@johnbumgardner Can you look at this Travis problem? I fixed auth_controller's protected method setup so that the methods would actually be protected, but now rspec can't test because they are protected. We can either do something like [this](https://stackoverflow.com/questions/267237/whats-the-best-way-to-unit-test-protected-private-methods-in-ruby) to test the protected methods, or just unprotect them in auth_controller. 

Since they were unprotected before and are now, it may change functionality. Haven't manually tested yet

johnbumgardner:
> @johnbumgardner Can you look at this Travis problem? I fixed auth_controller's protected method setup so that the methods would actually be protected, but now rspec can't test because they are protected. We can either do something like [this](https://stackoverflow.com/questions/267237/whats-the-best-way-to-unit-test-protected-private-methods-in-ruby) to test the protected methods, or just unprotect them in auth_controller.
> 
> Since they were unprotected before and are now, it may change functionality. Haven't manually tested yet

make it to where they arent protected at all. im not super concerned about encapsulation with these methods. Id rather just remove the protected keyword altogether.

nnhimes:
Did normal testing- made assignment with teams, did reviews, checked scores. All seems to work fine still

PR #2273, opened 2022-02-22
Fix the Rubric Delivering Order Problem

JesseChen1031:
The each_with_index iteration is moved to the above of <tr>


johnbumgardner (approved):
LGTM

PR #2277, opened 2022-02-23
Account request email fix

johnbumgardner:
Remove deprecated request user controller methods in users_controller.rb and include fullname of user in note sent to super administrator


nnhimes (approved):
Good find

PR #2282, opened 2022-02-26
issue fix #2281

### LINKED ISSUE #2281: Calibration report is wrong ... better to show #s of students than %s.
The Calibration Report purports to show the % of students that have chosen each answer, but the %s do not add to 100%:

image

Rather than fixing this bug, I would suggest just changing the report so it shows the _number_ of students who have chosen each answer.

johnbumgardner:
Make changes proposed in #2281


nnhimes (approved):
LGTM

PR #2286, opened 2022-03-03
fix issue #311 related to sending join requests to ones own team and disallowing users from graphically spamming invites

johnbumgardner:
added checks for if you were already on a team.
allow editing of invite rather than just sending a duplicate
closes issue #311


nnhimes (approved):
lgtm

nnhimes (changes requested):
See comment

PR #2295, opened 2022-03-09
E2200 Refactor advice_controller

AtharvaJ10:
Refactored and Tested the Advice Controller.
Refactored and Tested the Participants Helper.


app/controllers/advice_controller.rb
```diff
@@ -1,26 +1,34 @@
...
     @questionnaire = Questionnaire.find(params[:id])
 
+    # For each question in a quentionnaire, this method adjusts the advice size if the advice size is <,> nummber of advices or 
+    #the max or min score of the advices does not correspond to the max or min score of questionnaire respectively  
     @questionnaire.questions.each do |question|
-      num_questions = if question.is_a?(ScoredQuestion)
-                        @questionnaire.max_question_score - @questionnaire.min_question_score
+
+      #if the question is a scored question, store the number of advices corresponsing to that question (max_score - min_score), else 0
+      num_questions = if question.is_a?(ScoredQuestion)  #bug -> num_questions should be num_advices
```
namanshrimali, line 18: I concur. num_questions should be renamed to num_advices

app/controllers/advice_controller.rb
```diff
@@ -1,26 +1,34 @@
...
+                        @questionnaire.max_question_score - @questionnaire.min_question_score   #bug -> should add +1
```
namanshrimali, line 19: I concur, 1 should be added for calculating number of advices.

app/controllers/advice_controller.rb
```diff
@@ -1,26 +1,34 @@
...
+      #sorting question advices in descending order by score
       sorted_advice = question.question_advices.sort_by { |x| -x.score }
+
+      #Checks the condition for adjusting the advice size
       next unless (question.question_advices.length != num_questions) ||
                   sorted_advice.empty? ||
-                  (sorted_advice[0].score != @questionnaire.min_question_score) ||
-                  (sorted_advice[sorted_advice.length - 1] != @questionnaire.max_question_score)
+                  (sorted_advice[0].score != @questionnaire.min_question_score) || #bug -> should match with max_question_score 
+                  (sorted_advice[sorted_advice.length - 1] != @questionnaire.max_question_score) #bug -> should match with min_question_score
```
namanshrimali, line 32: This also looks good to me. 0th score should be compared with max_question_score and last score should be compared with min_question_score.

There's one change I would suggest. Please take all the conditions inside this unless condition and make a separate function, say `advices_matches?` (you can come up with a better name), which would return true or false based on these conditions.

app/controllers/advice_controller.rb
```diff
@@ -1,26 +1,34 @@
...
+
+      #if the question is a scored question, store the number of advices corresponsing to that question (max_score - min_score), else 0
+      num_questions = if question.is_a?(ScoredQuestion)  #bug -> num_questions should be num_advices
+                        @questionnaire.max_question_score - @questionnaire.min_question_score   #bug -> should add +1
                       else
                         0
                       end
 
+      #sorting question advices in descending order by score
       sorted_advice = question.question_advices.sort_by { |x| -x.score }
```
namanshrimali, line 25: Can you use `.reverse` for this ?
Something like this
```{ruby}
sorted_advice = question.question_advices.sort_by { |x| x.score }.reverse
```

app/controllers/advice_controller.rb
```diff
@@ -1,44 +1,60 @@
...
   include AuthorizationHelper
-
+  # If current user is TA then only current user can edit and update the advice
   def action_allowed?
     current_user_has_ta_privileges?
   end
 
+  # checks whether the advices for a question in questionnaire have valid attributes
+  # return true if the number of advices and their scores are invalid, else returns false
+  def invalid_advice_status(sorted_advice, num_advices, question)
```
namanshrimali, line 10: Can you look for better name for this method? [Ref1](https://stackoverflow.com/questions/3874350/naming-conventions-for-java-methods-that-return-boolean
), [Ref2](https://stackoverflow.com/questions/1263160/ruby-boolean-attribute-naming-convention-and-use
)

spec/controllers/advice_controller_spec.rb
```diff
@@ -22,4 +23,134 @@
       end
     end
   end
+
+  describe '#invalid_advice_status' do
+    context "when invalid_advice_status is called with question advice score > max score of questionnaire" do
+      #max score of advice = 3 (!=2)
+      let(:questionnaire) do
+        build(:questionnaire, id: 1, min_question_score: 1,
+          questions: [build(:question, id: 1, weight: 2, question_advices: [build(:question_advice, id:1, score: 1, question_id: 1, advice: "Advice1"), build(:question_advice, id:2, score: 3, question_id: 1, advice: "Advice2")])], max_question_score: 2)
```
namanshrimali, line 32: Can you break down these long lines of code into smaller lines, possibly break down the nested objects into separate variables?

spec/controllers/advice_controller_spec.rb
```diff
@@ -22,4 +23,134 @@
...
+        build(:questionnaire, id: 1, min_question_score: 1,
+          questions: [build(:question, id: 1, weight: 2, question_advices: [build(:question_advice, id:1, score: 1, question_id: 1, advice: "Advice1"), build(:question_advice, id:2, score: 2, question_id: 1, advice: "Advice2")])], max_question_score: 2)
+      end
+
+      it "edit advice redirects correctly when called" do
+        allow(Questionnaire).to receive(:find).with('1').and_return(questionnaire)
+        params = {id: 1}
+        session = {user: instructor1}
+        get :edit_advice, params, session
+        expect(response).to render_template(:edit_advice)
```
namanshrimali, line 123: Can you also check for response status after render/redirects?

spec/controllers/advice_controller_spec.rb
```diff
@@ -22,4 +23,134 @@
...
+      #max score of advice = 3 (!=2)
+      let(:questionnaire) do
+        build(:questionnaire, id: 1, min_question_score: 1,
+          questions: [build(:question, id: 1, weight: 2, question_advices: [build(:question_advice, id:1, score: 1, question_id: 1, advice: "Advice1"), build(:question_advice, id:2, score: 3, question_id: 1, advice: "Advice2")])], max_question_score: 2)
+      end
+
+      it "invalid_advice_status returns true when called with incorrect maximum score for a question advice" do
+        sorted_advice = questionnaire.questions[0].question_advices.sort_by { |x| x.score }.reverse
+        num_advices = questionnaire.max_question_score - questionnaire.min_question_score + 1  
+        temp = AdviceController.new
```
namanshrimali, line 38: Come up with a better name than `temp`

johnbumgardner: they dont have to initalize this object. rspec creates it for them, and it is saved locally with the variable name controller

app/controllers/advice_controller.rb
```diff
@@ -1,43 +1,60 @@
 class AdviceController < ApplicationController
```
namanshrimali, line 1: Can you add (as a code comment) what advice controller actually does in the overall application. You can basically copy and paste the content from your wiki page

PR #2301, opened 2022-03-10
Refactor create, action_allowed? and update methods in response controller

vamshi-chidara:
Modified code in create, action_allowed? and update methods of response controller


vamshi-chidara:
Created PR to main repo instead of forked repo

PR #2310, opened 2022-03-13
Downgraded bootstrap-sass version to 3.4.1

namanshrimali:
Currently expertiza sass is breaking, which is due to recent changes in bootstrap-sass npm package. Downgrading to last stable version (3.4.1) so that app runs without breaking and maintains styling.


johnbumgardner (approved):
lgtm

PR #2323, opened 2022-03-17
E2227/SQL injection fix

lobrien6588:
Changed to a dynamic finder in order to eliminate sql injection risks

@nnhimes

### DIFF
--- a/app/controllers/sign_up_sheet_controller.rb
+++ b/app/controllers/sign_up_sheet_controller.rb
@@ -422,7 +422,7 @@ class SignUpSheetController < ApplicationController
     team_id = TeamsUser.team_id(assignment.id, session[:user].id)
     original_topic_id = SignedUpTeam.topic_id(assignment.id.to_i, session[:user].id)
     #sql injection here
-    SignUpTopic.find(params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(params[:topic_id])
+    SignUpTopic.find_by(topic_id:params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(params[:topic_id])
     if SignedUpTeam.exists?(team_id: team_id, is_waitlisted: 0)
       SignedUpTeam.where(team_id: team_id, is_waitlisted: 0).first.update_attribute('topic_id', params[:topic_id].to_i)
     end
@@ -437,7 +437,7 @@ class SignUpSheetController < ApplicationController
 
   def publish_approved_suggested_topic
     #sql injection here
-    SignUpTopic.find(params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(params[:topic_id])
+    SignUpTopic.find(topic_id: params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(params[:topic_id])
     redirect_to action: 'list', id: params[:id]
   end
 
--- a/app/controllers/sign_up_sheet_controller.rb
+++ b/app/controllers/sign_up_sheet_controller.rb
@@ -421,6 +421,7 @@ class SignUpSheetController < ApplicationController
     assignment = AssignmentParticipant.find(params[:id]).assignment
     team_id = TeamsUser.team_id(assignment.id, session[:user].id)
     original_topic_id = SignedUpTeam.topic_id(assignment.id.to_i, session[:user].id)
+    #sql injection here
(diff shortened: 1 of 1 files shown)


nnhimes:
I enabled checks on your PR. The SQL injections are still there. You should be able to see if the Brakeman check passes or not now. You should remove that comment eventually since the injection won't exist anymore. I'd recommend trying to break the two lines up into multiple simpler lines. For example, find and assign the found data to an object. Then update the attributes.

spec/controllers/sign_up_sheet_controller_spec.rb
```diff
@@ -762,4 +762,24 @@
...
+      allow(SignedUpTeam).to receive(:where).with(team_id: 1, is_waitlisted: 0).and_return([signed_up_team])
+      allow(SignedUpTeam).to receive(:where).with(topic_id: 1, is_waitlisted: 1).and_return([signed_up_team])
+      allow(SignUpSheet).to receive(:signup_team).with(1, 8, 1).and_return('OK!')
+      params = {
+        id: 1,
+        topic_id: "' OR 1='1"
+      }
+      session = { user: instructor }
+      get :switch_original_topic_to_approved_suggested_topic, params, session
+      expect(response) # Change response
```
nnhimes, line 781: You need to have expect response to be equal to something or return nil. There are examples in other specs. Also, if you can please create another test case where something is returned.

nnhimes:
The current security output is a bit confusing:
Two SQL Injections were solved 3 days ago
image
image

But two injections remain in the open problem list:
image
Looking at these two open issues, they are the lines you fixed in the PR. I think once your PR is merged into beta these will be removed also.

app/controllers/sign_up_sheet_controller.rb
```diff
@@ -421,7 +421,7 @@ def switch_original_topic_to_approved_suggested_topic
     assignment = AssignmentParticipant.find(params[:id]).assignment
     team_id = TeamsUser.team_id(assignment.id, session[:user].id)
     original_topic_id = SignedUpTeam.topic_id(assignment.id.to_i, session[:user].id)
-    SignUpTopic.find(params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(params[:topic_id])
+    SignUpTopic.find_by(topic_id: params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(topic_id: params[:topic_id])
```
nnhimes, line 424: Are you sure topic_id is the right field to search by? Usually find() defaults to searching by that class' ID, and other files such as assessment360_controller.rb on line 131 have the find_by setup as:
image

spec/controllers/sign_up_sheet_controller_spec.rb
Woozl: How do you suggest we test a response that returns something? Like a generic input case that actually updates the topic?

app/controllers/sign_up_sheet_controller.rb
```diff
@@ -435,7 +435,7 @@ def switch_original_topic_to_approved_suggested_topic
   end
 
   def publish_approved_suggested_topic
-    SignUpTopic.find(params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(params[:topic_id])
+    SignUpTopic.find(topic_id: params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(topic_id: params[:topic_id])
```
nnhimes, line 438: See above comment

spec/controllers/sign_up_sheet_controller_spec.rb
nnhimes: Once a topic is updated it will redirect to another page I think. There are a lot of similar tests above this one in the same file.

app/controllers/sign_up_sheet_controller.rb
```diff
@@ -435,7 +435,7 @@ def switch_original_topic_to_approved_suggested_topic
   end
 
   def publish_approved_suggested_topic
-    SignUpTopic.find(params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(params[:topic_id])
+    SignUpTopic.find(id: params[:topic_id]).update_attribute(:private_to, nil) if SignUpTopic.exists?(id: params[:topic_id])
```
nnhimes, line 438: find_by, not find

lobrien6588: Long day

nnhimes:
That commit had gemfile changes

PR #2330, opened 2022-03-19
E2208: Testing for submission_records_controller, profile_controller

chirrag03:
- Added test cases for the methods action_allowed, index in the submission_records_controller
- IssueFix1 : check if user is the instructor before checking if the instructor instructs the course
- IssueFix2 : check if user is the TA before checking if the TA is assigned for the course

- Added test cases for the methods action_allowed, edit, update in the profile_controller

@nnhimes
@rachanasrib
@rshukla3


app/controllers/profile_controller.rb
```diff
@@ -31,6 +31,8 @@ def update
     redirect_to controller: :profile, action: :edit
   end
 
+  # :nocov:
+
```
nnhimes, line 35: Please remove this nocov

chirrag03: this is used for private method because as per "Rails Conf 2013 The Magic Tricks of Testing by Sandi Metz" taught in the coursework, we should not be testing private methods explicitly. And hence it should also not be included in the total coverage.

nnhimes:
Could your team please double-check that user_params is not used anywhere, and if not remove it from the controller?

PR #2340, opened 2022-03-20
E2215. Refactoring student_quizzes_controller.rb

gskn98:
Changes made: 
Line 29: .first has been changed to .last to count the score of the most recent attempt of the quiz.
Line 34: @participant has been changed to @quiz_taker to more accurately describe its function. This change has also been reflected in finished_quiz.html.erb.
Line 39: The method comment of self.take_quiz has been updated to describe its function
-Controller Method calculate_score has been split into three separate functions and has been moved to student_quizzes_helper.rb.
Line 60: A method comment has been added for record_response to describe its function
Line 72: The Flash message has been updated.
-graded? has been removed as it is no longer necessary.
Line 77: The method comment for review_questions has been updated.


johnbumgardner:
thinking this not be merged and we remove quiz functionality from expertiza

PR #2357, opened 2022-03-28
Reputation hamer testing

joshlin5:
Added two rspec test files to test the reputation web server hamer algorithm:
expertiza/spec/controllers/reputation_web_server_hamer.rb
expertiza/spec/controllers/reputation_mock_web_server_hamer.rb
@efg is the mentor for this team.


joshlin5:
Two commits fail the Travis CI because of the commit: "added gem webmock and net-http". However, in the latest commit, we fixed the mistakes from that commit by removing those gems. So the latest commit passes with no issues.

PR #2368, opened 2022-04-08
Fix Tag Prompts

nnhimes:
From issue #2359 
Can add tag prompts again, and also changes the spelling for the tag prompt toggle button to be clearer.

PR #2370, opened 2022-04-08
Rubocop Fix

nnhimes:
Cleaning up the thousands of rubocop errors after Rails 5 update


nnhimes:
The last rubocop error will be fixed in #2368

PR #2373, opened 2022-04-14
Fix Crash for No Team Submissions

nnhimes:
Just a temporary fix for the crash when submitting something on an assignment without teams. So many functions need the participant to be on a team, and submitting content when not on a team is literally not possible with how Expertiza is setup. 

I think we eventually need to replace the option to enable teams with a forced box that defaults to 2 members per team. If you decrease it to one, that would accomplish what we want while not breaking everything. 

image

image


johnbumgardner (approved):
lgtm

PR #2379, opened 2022-04-19
Force team creation

nnhimes:
Expertiza does not work when teams are disabled and students are by themselves. This forces team creation with a default of 1 person teams.

Tested submitting links and files, reviewing, and viewing scores/reports.

All seems to be fine, the two issues I found exist on beta for bigger teams as well I think.

Issue #2258 :
I submitted a review from two students on each other. The bar chart shows the scores, but only one team appears in the bottom table.
image

Issue 2:
Submitting a grade/comment does not save from Review Report. This one may be a new issue?
image

PR #2409, opened 2022-05-01
Fix forced team creation

nnhimes:
When making an assignment with more than 1 person teams, the teammate review row in rubrics is not added.

The hasTeamsChanged() method used to call a function to do this. I removed hasTeamsChanged though because it used to be called on checkbox change, and all of its other functionality besides calling addDueDateTableElement() is useless now.

I tried adding the call to addDueDateTableElement() at the bottom in the jQuery(document).ready(function() {} section, like how it is in other files, but it doesn't seem to work still

### DIFF
--- a/app/views/assignments/edit/_general.html.erb
+++ b/app/views/assignments/edit/_general.html.erb
@@ -94,7 +94,7 @@ function autogenerate_submission(){
     <td style='padding:5px' id='assignment_team_assignment_field'>
 
       <div class="form-inline">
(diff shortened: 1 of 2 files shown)


nnhimes:
@johnbumgardner Can you look at this and the addDueDateTableElement() call?

johnbumgardner:
so PR https://github.com/expertiza/expertiza/pull/2404 actually touches the due date code. maybe check to see if this PR fixes the issue then try rewriting.

nnhimes:
Reverting the force team creation code from #2379. PR #2404 needs to be merged and changes method locations that make everything complicated. 

**A future project needs to finish what #2379 and #2409 try to do.**

PR #2410, opened 2022-05-01
Fix back buttons

### LINKED ISSUE #2369: Bad Back Buttons
Many back buttons redirect back into forms instead of where they should go. I'm starting a list here and will try to fix soon

- Other's Work in Student Assignment View. Goes into form after doing review, should go back to assignment view page.
- Your Scores in Student Assignment View. Has no back button.
- After doing teammate review on student team, goes back into form. Should go back to assignment view page.

nnhimes:
Fixes #2369 - three of the worst back button problems


johnbumgardner (approved):
LGTM

PR #2421, opened 2022-05-10
Classes & assignments deletions

### LINKED ISSUE #2420: Can't Delete Assignments or Courses
When trying to delete assignment (no console error): 
image

When trying to delete course:
VERY OLD COURSE (WCAE 2008):
image
image
NEWER COURSE (506 Summer 2020):
image
image

nnhimes:
Fixes #2420 

Tested by making a course, making an assignment in it. Created teams, did reviews, scores all work. Then I deleted the entire course and the assignment also disappeared. Deleting individual assignments also works.

### DIFF
--- a/app/models/response_map.rb
+++ b/app/models/response_map.rb
@@ -1,7 +1,7 @@
 class ResponseMap < ApplicationRecord
   extend Scoring
   has_many :response, foreign_key: 'map_id', dependent: :destroy, inverse_of: false
-  belongs_to :reviewer, class_name: 'Participant', foreign_key: 'reviewer_id', inverse_of: false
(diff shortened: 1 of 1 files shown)


nnhimes:
Not sure if this affects database, I think it just adds the relational call so the inverse association can be found

johnbumgardner (approved):
LGTM. Does this generally fix the course deletion issue?

nnhimes:
@johnbumgardner Yeah I think the only thing stopping courses from being deleted before was the error with the response map/assignment relation when it would delete assignments belonging to that course. So since assignments can be found for the response map now, it can delete assignments and then delete the course

Issue #2412, opened 2022-05-03
Non-team Assignments Broken


nnhimes:
Expertiza does not work when teams are disabled and students are by themselves. So many methods essential to Expertiza do not work without teams enabled. Students cannot submit even submit work.

I created two PRs (now closed because of Spring 22 final project complications) that force team creation in Expertiza and allow 1 person teams. 
#2379 Made teams of as little of 1 person be made, and students on a team by themselves can submit and review work.
#2409 Tried fixing an issue with the first PR. When making an assignment with more than 1 person teams, the teammate review row in rubrics is not added. It also forced being on a team before clicking the "Your Work" button. There was an issue with addDueDateTableElement() that was not solved yet.

A future project should use these and finish the work.

PR #2422, opened 2022-05-13
Force teams

nnhimes:
Follows on issue #2412 and PRs #2379 and #2409.


nnhimes:
@johnbumgardner I don't know why it's not working. I'm calling a new method I made from the onChange option of the team size field. The puts I added show its being called when it should be, but the if statement inside is always true for some reason (the puts show the values change and are integers). And inside that if statement it calls the addDueDateTableElement() method that the old code was calling. The table is just never updated with the new row for some reason. 

image

johnbumgardner (approved):
LGTM now Nicholas

PR #2472, opened 2022-10-27
E2257 Refactor questionnaires_controller.rb

Ashwinshankar98:
1. Removed create_questionnaire
2. Split create method into smaller self-documenting methods.
3. QuizQuestionnaire check
4. Refactor add_new_questions method
5. Remove redundant variables in QuestionnairesController
6. Code climate fixes

Wiki - [https://expertiza.csc.ncsu.edu/index.php/CSC/ECE_517_Fall_2022_-_E2257:_Refactor_questionnaires_controller.rb](https://expertiza.csc.ncsu.edu/index.php/CSC/ECE_517_Fall_2022_-_E2257:_Refactor_questionnaires_controller.rb)


app/models/question.rb
```diff
@@ -31,6 +31,10 @@ class Question < ApplicationRecord
   ANSWERS_TRUE_FALSE = [['1', 1], ['2', 2]].freeze
```
efg, line 31: I don't understand what these constants are for.  I presume they are defaults, but their purpose should be explained by comment(s).

PR #2484, opened 2022-11-28
Add participant_id to teams_users table

SahithiAmmana:
### E2283. Refactor student_teams functionality

- Created a column that would be called “participant_id”, which will be a foreign key that references participants table.
- Created a migration that fetches the assignment_id for each tuple from the teams table. We would find the participant with the help of the assignment_id we just got from the teams table and the user_id that is already present in the current table, and store its “id” in the participant_id column we created in the last step. You have to come with optimal migration strategy that reliably calculates the participant_id

@rjain09
@namanshrimali 

About the Expertiza Bot
- If you have any questions about comments given by the expertiza-bot [(example)](https://github.com/expertiza/expertiza/pull/1877#issuecomment-762412918), you could create a new comment in your pull request with the content `/dispute [UUID1] [UUID2]` [(example)](https://github.com/expertiza/expertiza/pull/1877#issuecomment-762430057), then the professor and TAs will be notified.
- [Here is a video](https://tinyurl.com/internet-bots) about how to communicate to the expertiza-bot.

### DIFF
--- a/db/migrate/20221128192313_add_participant_id_to_teams_users.rb
+++ b/db/migrate/20221128192313_add_participant_id_to_teams_users.rb
@@ -1,0 +1,22 @@
+class AddParticipantIdToTeamsUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :teams_users, :participant_id, :integer, index: true
+    # add_reference :teams_users, :participants
+    add_foreign_key :teams_users, :participants
+
+    teams_users = TeamsUser.all
+    teams_users.each do |teams_user|
+      if !teams_user.team_id.nil?
+        team = Team.find(teams_user.team_id)
+        if !team.nil?
+          participant = Participant.find_by(user_id: teams_user["user_id"], parent_id: team["parent_id"])
+          if !participant.nil?
+            teams_user.participant_id = participant["id"]
+            teams_user.save
+          end
+        end
+      end
+      next unless teams_user.participant_id
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20220405222420) do
+ActiveRecord::Schema.define(version: 20221128192313) do
 
   create_table "account_requests", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
     t.string "name"
@@ -718,7 +718,9 @@ ActiveRecord::Schema.define(version: 20220405222420) do
     t.integer "user_id"
     t.integer "duty_id"
     t.string "pair_programming_status", limit: 1
+    t.integer "participant_id"
     t.index ["duty_id"], name: "index_teams_users_on_duty_id"
+    t.index ["participant_id"], name: "fk_rails_7192605c92"
     t.index ["team_id"], name: "fk_users_teams"
     t.index ["user_id"], name: "fk_teams_users"
   end
@@ -832,6 +834,7 @@ ActiveRecord::Schema.define(version: 20220405222420) do
   add_foreign_key "tag_prompt_deployments", "questionnaires"
   add_foreign_key "tag_prompt_deployments", "tag_prompts"
   add_foreign_key "teams_users", "duties"
+  add_foreign_key "teams_users", "participants"
   add_foreign_key "teams_users", "teams", name: "fk_users_teams"
   add_foreign_key "teams_users", "users", name: "fk_teams_users"
 end


db/migrate/20221128192313_add_participant_id_to_teams_users.rb
```diff
@@ -0,0 +1,22 @@
+class AddParticipantIdToTeamsUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :teams_users, :participant_id, :integer, index: true
+    # add_reference :teams_users, :participants
+    add_foreign_key :teams_users, :participants
+
+    teams_users = TeamsUser.all
+    teams_users.each do |teams_user|
+      if !teams_user.team_id.nil?
```
namanshrimali, line 9: Use `unless` instead of ! with nil

db/migrate/20221128192313_add_participant_id_to_teams_users.rb
```diff
@@ -0,0 +1,22 @@
...
+        team = Team.find(teams_user.team_id)
+        if !team.nil?
```
namanshrimali, line 11: use unless

db/migrate/20221128192313_add_participant_id_to_teams_users.rb
```diff
@@ -0,0 +1,22 @@
...
+          participant = Participant.find_by(user_id: teams_user["user_id"], parent_id: team["parent_id"])
+          if !participant.nil?
```
namanshrimali, line 13: use unless

db/migrate/20221128192313_add_participant_id_to_teams_users.rb
```diff
@@ -0,0 +1,22 @@
...
+            teams_user.participant_id = participant["id"]
+            teams_user.save
+          end
+        end
+      end
+      next unless teams_user.participant_id
```
namanshrimali, line 19: I'm not sure what it does

app/controllers/student_teams_controller.rb
```diff
@@ -32,7 +32,9 @@ def action_allowed?
     when 'create'
       current_user_has_id? student.user_id
     when 'edit', 'update'
-      current_user_has_id? team.user_id
+      if @team!=nil
```
namanshrimali, line 35: This should not be the case, but I'll check on it and see how I can resolve this

app/models/course_team.rb
```diff
@@ -65,9 +65,20 @@ def self.export_fields(options)
...
+    add_participant(parent_id, user)
+    participant = CourseParticipant.find_by(parent_id: parent_id, user_id: user.id)
+    t_user = TeamsUser.create(participant_id: participant.id, team_id: id)
+    parent = TeamNode.find_by(node_object_id: id)
+    TeamUserNode.create(parent_id: parent.id, node_object_id: t_user.id)
+  end
+
+  # Add participant to a team.
+  # Raise exception if the participant is already part of this team.
+  def add_participant_to_team(participant, _id = nil)
```
namanshrimali, line 77: Can you send me a screenshot that this works?

app/models/course_team.rb
```diff
@@ -65,9 +65,20 @@ def self.export_fields(options)
   def add_member(user, _id = nil)
     raise "The user \"#{user.name}\" is already a member of the team, \"#{name}\"" if user?(user)
 
-    t_user = TeamsUser.create(user_id: user.id, team_id: id)
+    add_participant(parent_id, user)
```
namanshrimali, line 68: This code should not be present. If we have participant_id in our table, we should directly get the participant's information and not through user_id.

app/models/course_team.rb
```diff
@@ -65,9 +65,20 @@ def self.export_fields(options)
...
+    participant = CourseParticipant.find_by(parent_id: parent_id, user_id: user.id)
+    t_user = TeamsUser.create(participant_id: participant.id, team_id: id)
+    parent = TeamNode.find_by(node_object_id: id)
+    TeamUserNode.create(parent_id: parent.id, node_object_id: t_user.id)
```
namanshrimali, line 72: Again, why are we using user_id?

app/models/team.rb
```diff
@@ -5,15 +5,30 @@ class Team < ApplicationRecord
...
+  # have user_id mapping. This function loads participants details directly from the mapping
+  # as well as from user_id if some entries use user_id mapping.
   # Get the participants of the given team
   def participants
-    users.where(parent_id: parent_id || current_user_id).flat_map(&:participants)
+    # load directly using participant_ids
+    participants_list_from_ids = Participant.where(id: participant_ids)
+
+    unless participants_list_from_ids.nil?
+      participants_list = []
```
namanshrimali, line 28: Can you use [append method](https://www.geeksforgeeks.org/ruby-array-append-function/)

app/models/team.rb
```diff
@@ -5,15 +5,30 @@ class Team < ApplicationRecord
...
+      participants_list += participants_list_from_ids unless participants_list_from_ids.nil?
```
namanshrimali, line 29: May never be nil. Try using empty?

app/models/team.rb
```diff
@@ -65,15 +88,32 @@ def add_member(user, _assignment_id = nil)
     can_add_member = false
     unless full?
       can_add_member = true
-      t_user = TeamsUser.create(user_id: user.id, team_id: id)
       parent = TeamNode.find_by(node_object_id: id)
-      TeamUserNode.create(parent_id: parent.id, node_object_id: t_user.id)
       add_participant(parent_id, user)
+      participant = AssignmentParticipant.find_by(parent_id: parent_id, user_id: user.id)
+      t_user = TeamsUser.create(participant_id: participant.id, team_id: id)
+      TeamUserNode.create(parent_id: parent.id, node_object_id: t_user.id)
```
namanshrimali, line 95: Can you let me know what TeamUserNode does? How can we use participant_id here?

app/views/student_teams/_team_display.html.erb
```diff
@@ -0,0 +1,71 @@
...
+        <th class="head">Email address</th>
+
+        <% if @teammate_review_allowed %>
+            <th class="head"><%=t ".review_action"%></th>
+        <% end %>
+    </tr>
+
+    <% @student.team.participants.each do |member| %> <!--start team member content-->
+        <tr>
+            <td><%= member.user.name(session[:ip]) %></td>
```
namanshrimali, line 17: Information of the user should be fetched through participant_id.

db/migrate/20221128192313_add_participant_id_to_teams_users.rb
SahithiAmmana: Changed it to `unless`

db/migrate/20221128192313_add_participant_id_to_teams_users.rb
SahithiAmmana: Changed it to `unless`

db/migrate/20221128192313_add_participant_id_to_teams_users.rb
SahithiAmmana: Changed it to `unless`

db/migrate/20221128192313_add_participant_id_to_teams_users.rb
SahithiAmmana: It is a condition to check if the teams_user.participant_id is not nil before proceeding with the next user in teams_users table.

app/models/course_team.rb
SahithiAmmana: We're finding participant id based on the id from Users table and assignment id, the creating a new TeamsUser. To display in the team in tree structure, TeamUserNode is created with **id** from teams_users table

app/models/team.rb
SahithiAmmana: To display in the team in tree structure, TeamUserNode is created with id from teams_users table

app/models/team.rb
SahithiAmmana: `append` is not working. Used `push` to append the array

app/views/student_teams/_team_display.html.erb
SahithiAmmana: Updated 👍

PR #2538, opened 2023-03-17
[WIP] Dd main bug fixes

divyang02:
fixed vary by round


app/models/assignment.rb
```diff
@@ -320,7 +320,7 @@ def current_stage_name(topic_id = nil)
   # check if this assignment has multiple review phases with different review rubrics
   def varying_rubrics_by_round?
     # E-2084 corrected '>=' to '>' to fix logic
```
divyang02, line 322: Put a comment - "This is a hack, we should actually check if we have more than one rubric of a given type eg, review

PR #2640, opened 2023-10-01
Fixing the lint workflow to ignore certain files which do not need to be spell checked

qureshi-ali:
**What has been changed: ** Ignored a pdf file as well as the db/migrate folder in the lint.yml workflow to fix the lint runner.


qureshi-ali:
Closed as earlier commits were not supposed to be part of the PR

PR #2658, opened 2023-10-18
E2352. Fix "Back" link on “New Late Policy” page

Sarvesh-Somasundaram:
- no changes have been made, this is just the initial PR

### DIFF
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -47,6 +47,7 @@ class GradesController < ApplicationController
       end
     end
     @scores = review_grades(@assignment, @questions)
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -26,4 +26,18 @@ def flash_message(type)
   def text_field_with_auto_complete(model, field, options)
     text_field_tag "#{model}[#{field}]", '', options
   end
+
+  def custom_error_messages_for(object)
```
efg, line 30: This method badly needs a method comment.  Why are there "messages" (plural)?

PR #2692, opened 2023-11-14
Fixing bug where instructors were unable to see all assignment teams when viewing scores.

qureshi-ali:
**What has been changed:** 

When looking at participant scores under the view scores section, the instructors could not see all the participants in the assignment. This was occurring due to index of the hash not changing after rejecting nil scores. 

A new variable that keeps track of the original length of the scores before rejecting the nil scores has been added. This makes sure that all participants are iterated over when displaying the teams.

Also fixed bug where instructors where not being able to export grades to csv. This was because the code was calling the review_grades method in a class method. An instance of assignment had to be created to call it.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -500,7 +500,8 @@ class Assignment < ApplicationRecord
       end
       @questions[questionnaire_symbol] = questionnaire.questions
     end
-    @scores = review_grades(self, @questions)
+    assignment_instance = Assignment.new
+    @scores = assignment_instance.review_grades(@assignment, @questions)
     return csv if @scores[:teams].nil?
 
     export_data(csv, @scores, options)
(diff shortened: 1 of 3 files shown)


ameyagv (approved):

app/models/assignment.rb
```diff
@@ -500,7 +500,8 @@ def self.export(csv, parent_id, options)
       end
       @questions[questionnaire_symbol] = questionnaire.questions
     end
-    @scores = review_grades(self, @questions)
+    assignment_instance = Assignment.new
+    @scores = assignment_instance.review_grades(@assignment, @questions)
```
ameyagv, line 504: can we remove assignment_instance? and call without instance variable

app/models/assignment.rb
```diff
@@ -510,38 +511,41 @@ def self.export_data(csv, scores, options)
     @scores = scores
     (0..@scores[:teams].length - 1).each do |index|
       team = @scores[:teams][index.to_s.to_sym]
+      first_participant = team[:team].participants[0] unless team[:team].participants[0].nil?
```
ameyagv, line 514: write a method or search for a method that checks whether the team is empty. Team.size() can be used?

app/models/assignment.rb
```diff
@@ -510,38 +511,41 @@ def self.export_data(csv, scores, options)
...
+      next if first_participant.nil?
+      pscore = @scores[:participants][first_participant.id.to_s.to_sym]
```
ameyagv, line 516: rename pscore to participant_score

PR #2769, opened 2024-03-24
E2403. Mentor-meeting management [Emails Notifications]

ExtremeMachine12:
Changes Proposed in Pull Request:

Added a new feature such that users are automatically emailed when they are added to a team. Also, mentors are automatically emailed a different email when they are added to a team. 
Added a new function to the mailer_helper class for mail specifically sent to user/mentors when added to a team. 
Added view partials to handle the HTML format of the new emails being sent. 
Slightly changed the generic_message function in the mailer class to handle more cases. 
Also added test cases to check that users/mentors are emailed when added to teams now. 

Mentions:
@efg

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -89,7 +89,10 @@ class Team < ApplicationRecord
       add_participant(parent_id, user)
       ExpertizaLogger.info LoggerMessage.new('Model:Team', user.name, "Added member to the team #{id}")
       # Now that a new team member has been added to a team, send an email to them letting them know
-      MailerHelper.send_team_confirmation_mail_to_user(user, "You have been added to the team #{name}.", "team_welcome").deliver
+      MailerHelper.send_team_confirmation_mail_to_user(user, "[Expertiza] Added to a Team", "user_added_to_team", "#{name}", Assignment.find(_assignment_id).name.to_s).deliver
+      #Mailer.generic_message(to: user,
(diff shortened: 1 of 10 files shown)


app/models/team.rb
```diff
@@ -88,6 +88,17 @@ def add_member(user, _assignment_id = nil)
       TeamUserNode.create(parent_id: parent.id, node_object_id: t_user.id)
       add_participant(parent_id, user)
       ExpertizaLogger.info LoggerMessage.new('Model:Team', user.name, "Added member to the team #{id}")
+      # if assignment_id is nil, then don't send an assignment name
+      assignment_name = assignment_id ? Assignment.find(assignment_id).name.to_s : ''
+      # Now that a new team member has been added to a team, send an email to them letting them know
+      if MentorManagement.user_a_mentor?(user)
+        MailerHelper.send_team_confirmation_mail_to_user(user, '[Expertiza] Added to a Team', 'mentor_added_to_team', name.to_s, assignment_name).deliver
+      elsif !user.is_a?(Participant)
```
efg, line 96: But a mentor is also a participant in the assignment.  I don't understand this line.

efg (commented):
It seems that this is only some of your changes.  You submitted another pull request, but that one changed more than 1000 files!  So I don't know where to find the full set of your changes.

PR #2774, opened 2024-03-25
E2414 - Grading audit trail

TylerEnck:
Implements a feature to maintain a history of all changes made to grades for assignments and reviews. When an instructor enters a grade for an assignment or review, they have the option to view the grading history, which will display a listing of all of the grades and comments previously given.

This implementation is based on on work from E2383, adding several UI formatting changes, refactoring to improve coding style, and additional controller tests


app/controllers/grading_histories_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  # return all grading history entries for the assignment
+  # entries are returned in chronological order
+  def index
+    @grading_histories = GradingHistory.where(graded_member_id: params[:graded_member_id], graded_item_type: params[:grade_type]).reverse_order
+    record = @grading_histories[0]
+    if record == nil
+      @receiver = ""
+      @assignment = ""
+    else
```
efg, line 32: This line needs a comment saying what the else condition signifies.

app/models/grading_history.rb
```diff
@@ -0,0 +1,20 @@
+class GradingHistory < ActiveRecord::Base
+  belongs_to :instructor
+  belongs_to :assignment
+
+  # populate the assignment fields according to type
+  def self.assignment_for_history(type, graded_member_id, participant_id)
+    # for a submission, the receiver is an AssignmentTeam
```
efg, line 7: It's a class method, so the receiver is the class, no?

app/models/review_grading_history.rb
```diff
@@ -0,0 +1,4 @@
+class ReviewGradingHistory < GradingHistory
+  attr_protected
+  belongs_to :graded_member, class_name: 'Participant'
+end
```
efg, line 4: Please add newlines at the end of files to avoid these annoying markers.

efg (commented):
Looks good!

PR #2793, opened 2024-04-21
Sudama

sagar110599:
1) Renamed Variables who_user and renamed functions
2) Fix to the codeql error raised in build


sagar110599:
Redudant Branch

PR #2867, opened 2024-10-27
find_teammates_by_user static method moved to helper file

moreharsh:
1. find_teammates_by_user static method moved to helper file
2. code updated as per code climate requirements


moreharsh:
Duplicate

PR #2913, opened 2025-02-04
E2452. Refactor review_mapping_controller.rb

Anish7-anish:
Renamed Methods:
I updated various method names in review_mapping_controller.rb and review_mapping_helper.rb to be clearer and more in line with Ruby conventions. For example, get_participant_or_reviewer is now find_assignment_participant, and check_for_self_review? is now is_self_review?.

### DIFF
--- a/app/controllers/review_mapping_controller.rb
+++ b/app/controllers/review_mapping_controller.rb
@@ -25,11 +25,11 @@ class ReviewMappingController < ApplicationController
   end
 
   def add_calibration
-    participant = get_participant_or_reviewer(session[:user].id, params[:id])
+    participant = find_assignment_participant(session[:user].id, params[:id])
     if participant.nil?
       participant = AssignmentParticipant.create(parent_id: params[:id], user_id: session[:user].id, can_submit: 1, can_review: 1, can_take_quiz: 1, handle: 'handle')
     end
-    map = get_review_response_mapping(params[:id], participant, params[:team_id], true)
+    map = find_review_response_mapping(params[:id], participant, params[:team_id], true)
 
     if map.nil?
       map = ReviewResponseMap.create(reviewed_object_id: params[:id], reviewer_id: participant.get_reviewer.id, reviewee_id: params[:team_id], calibrate_to: true) if map.nil?
@@ -47,12 +47,12 @@ class ReviewMappingController < ApplicationController
     @mapping = ResponseMap.find(params[:id])
   end
 
-  
+
   #EDITED: If the user is invalid(not present in database), then return with error message.
   #If reviewer is a participant of the assignment, then only move forward with reviewer assignment.
-  #Work Done: Refactored add_reviewer and extracted some code to make it smaller. 
+  #Work Done: Refactored add_reviewer and extracted some code to make it smaller.
   def add_reviewer
-    assignment = get_assignment(params[:id])
+    assignment = find_assignment(params[:id])
     topic_id = params[:topic_id]
     user_id = if User.where(name: params[:user][:name]).first
       then User.where(name: params[:user][:name]).first.id
@@ -65,7 +65,7 @@ class ReviewMappingController < ApplicationController
     end
     # If instructor wants to assign one student to review his/her own artifact,
     # it should be counted as "self-review" and we need to make /app/views/submitted_content/_selfreview.html.erb work.
-    if check_for_self_review?(params[:contributor_id], user_id)
+    if is_self_review?(params[:contributor_id], user_id)
       flash[:error] = 'You cannot assign this student to review his/her own artifact.'
       redirect_to action: 'list_mappings', id: assignment.id
       return
@@ -80,7 +80,7 @@ class ReviewMappingController < ApplicationController
     rescue StandardError => e
       msg = e.message
     end
-  
+
     redirect_to action: 'list_mappings', id: assignment.id, msg: msg
   end
   def add_metareviewer
@@ -138,13 +138,13 @@ class ReviewMappingController < ApplicationController
   # @raise [RuntimeError] Raises an error if the reviewer is already assigned to the contributor.
   # @return [ReviewResponseMap] The newly created review response mapping record.
   def assign_reviewer_to_contributor(user, assignment, contributor_id, topic_id)
-  
+
     # Generate registration URL
     reg_url = url_for id: assignment.id, user_id: user.id, contributor_id: contributor_id
-  
+
     # Get the assignment's participant corresponding to the user
     reviewer = get_reviewer(user, assignment, reg_url)
-  
+
     # Check if reviewer is already assigned
     if ReviewResponseMap.exists?(reviewee_id: contributor_id, reviewer_id: reviewer.id)
       raise "The reviewer, \"#{reviewer.name}\", is already assigned to this contributor."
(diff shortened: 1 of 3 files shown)


app/controllers/review_mapping_controller.rb
```diff
@@ -25,23 +25,19 @@ def action_allowed?
   end
 
   def add_calibration
-    participant = begin
-                    AssignmentParticipant.where(parent_id: params[:id], user_id: session[:user].id).first
-                  rescue StandardError
-                    nil
-                  end
+    participant = find_assignment_participant(session[:user].id, params[:id])
```
efg, line 28: Why was the rescue clause removed?

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
     @mapping = ResponseMap.find(params[:id])
   end
 
+
+  #EDITED: If the user is invalid(not present in database), then return with error message.
```
efg, line 51: Please make sure to skip a space before all open parens that do nob begin parameter lists.

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
...
+  #If reviewer is a participant of the assignment, then only move forward with reviewer assignment.
```
efg, line 52: Please reword this comment.  Should it be something like, "A user may only be assigned to review someone if that reviewer is a participant in the assignment."?  But it still is not very clear which lines of code enforce this provision.

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
...
+  #Work Done: Refactored add_reviewer and extracted some code to make it smaller.
   def add_reviewer
-    assignment = Assignment.find(params[:id])
+    assignment = find_assignment(params[:id])
     topic_id = params[:topic_id]
-    user_id = User.where(name: params[:user][:name]).first.id
-    # If instructor want to assign one student to review his/her own artifact,
+    user_id = if User.where(name: params[:user][:name]).first
```
efg, line 57: Lines 57-59 are unclear and need a comment explaining what they do.

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
...
+      then User.where(name: params[:user][:name]).first.id
+      else nil
+      end
+    if user_id.nil?
+      flash[:error] = 'The user does not exist.'
+      redirect_to action: 'list_mappings', id: assignment.id, msg: 'The user does not exist.'
+      return
+    end
+    # If instructor wants to assign one student to review his/her own artifact,
     # it should be counted as "self-review" and we need to make /app/views/submitted_content/_selfreview.html.erb work.
```
efg, line 67: This comment appears to say that the instructor can assign a self-review, but the code says he can't.  Please explain.

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
...
-    if TeamsUser.exists?(team_id: params[:contributor_id], user_id: user_id)
+    if is_self_review?(params[:contributor_id], user_id)
       flash[:error] = 'You cannot assign this student to review his/her own artifact.'
+      redirect_to action: 'list_mappings', id: assignment.id
+      return
+    end
+    # Team lazy initialization
```
efg, line 73: How is this lazy initialization?  Of the team?  So, a team is not created until necessary.  I understand that, but what code here creates a team?!

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
...
+      user = User.from_params(params)
+      # Assign the reviewer to the contributor
+      assign_reviewer_to_contributor(user, assignment, params[:contributor_id], topic_id)
+    rescue StandardError => e
+      msg = e.message
+    end
+
+    redirect_to action: 'list_mappings', id: assignment.id, msg: msg
+  end
+  def add_metareviewer
```
efg, line 86: Since metareviews have not been used since 2011, I wonder why you added code to implement them.  Or, was it just moved from somewhere else?  Where?  Why?

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
...
+
+      MetareviewResponseMap.create(reviewed_object_id: mapping.map_id,
+                                   reviewer_id: reviewer.id,
+                                   reviewee_id: mapping.reviewer.id)
+    rescue StandardError => e
+      msg = e.message
+    end
+    redirect_to action: 'list_mappings', id: mapping.assignment.id, msg: msg
+  end
+
```
efg, line 106: This is an awful lot of new code.  I suspect it came from somewhere else.  Why was it moved?

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
...
-    end
-  end
+
 
   # assigns the quiz dynamically to the participant
   def assign_quiz_dynamically
     begin
-      assignment = Assignment.find(params[:assignment_id])
-      reviewer = AssignmentParticipant.where(user_id: params[:reviewer_id], parent_id: assignment.id).first
+      assignment = find_assignment(params[:assignment_id])
```
efg, line 238: Why is a method call being substituted for a direct invocation of a db function?  This seems to make the code more obscure instead of more readable.  I think this should be reverted.

app/controllers/review_mapping_controller.rb
```diff
@@ -51,129 +47,196 @@ def select_metareviewer
...
+      reviewer = find_assignment_participant(params[:reviewer_id], assignment.id)
```
efg, line 239: Ditto for this line.

app/helpers/review_mapping_helper.rb
```diff
@@ -2,7 +2,175 @@ module ReviewMappingHelper
...
+#ADDED: Get participant details if user is added to the assignment
+#RENAMED: get_participant_or_reviewer -> find_assignment_participant
+  # @param user_id [Integer] The ID of the user.
+  # @param parent_id [Integer] The ID of the assignment.
+  # @return [AssignmentParticipant or nil] The participant record if found, or nil if not found.
+
+# Reasoning:
+  #  - Reflects that this method locates an AssignmentParticipant for a given user and assignment.
+  #  - Removes ambiguity around "or reviewer," since it specifically returns a participant.
+  def find_assignment_participant(user_id, parent_id)
```
efg, line 14: Why not just remove this method and call the db query wherever it's needed?  That would seem to be a much clearer way to code.

spec/controllers/review_mapping_controller_spec.rb
```diff
@@ -682,4 +730,38 @@
...
+      expect(assigns(:contributor)).to eq(team)
+    end
+
+    it 'calls AssignmentTeam.find with the correct contributor_id' do
+      expect(AssignmentTeam).to have_received(:find).with('1')
+    end
+  end
+
 end
+
```
efg, line 767: Please add an EoL to suppress the ugly warning.

efg (changes requested):

